PR #98, opened 2022-04-17
set AH in int10 handler

akaWolf:


maximilien-noal:
Thank you for this PR.

I'm curious, what does this fix ?

akaWolf:
> I'm curious, what does this fix ?

I'm reversing The Lost Vikings, it checks for AX value after interrupt

PR #121, opened 2022-05-18
Fix/gdb leaks and small stuff

maximilien-noal:
* Fixes ManualResetEvents not being disposed (possible resource leaks)
* PauseHandler uses a ManualResetEvent to pause/resume the running thread instead of a while loop and Monitor + lock.
* GdbServer uses a named Thread instead of a BackgroundWorker (less code, more easily debuggable)
* More logs on what is sent to GDB

Tested agaisnt master and the java version successfully :

- gdb startup
- -continue command
- gdb breakpoint from the Pause button from Spice86

### DIFF
--- a/src/Spice86/Emulator/Gdb/GdbCommandHandler.cs
+++ b/src/Spice86/Emulator/Gdb/GdbCommandHandler.cs
@@ -65,6 +65,9 @@ public class GdbCommandHandler {
                 'Z' => _gdbCommandBreakpointHandler.AddBreakpoint(commandContent),
                 _ => _gdbIo.GenerateUnsupportedResponse()
             };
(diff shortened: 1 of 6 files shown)


src/Spice86/Emulator/Gdb/GdbCommandHandler.cs
```diff
@@ -65,6 +65,9 @@ public class GdbCommandHandler {
                 'Z' => _gdbCommandBreakpointHandler.AddBreakpoint(commandContent),
                 _ => _gdbIo.GenerateUnsupportedResponse()
             };
+            if (_logger.IsEnabled(Serilog.Events.LogEventLevel.Information)) {
+                _logger.Information("Responded with {@Response}", response);
```
kevinferrare, line 69: This is already supposed to be logged in _gdbIo.SendResponse

PR #157, opened 2022-08-31
Sanitize Pic behaviour

kevinferrare:


src/Spice86.Core/Emulator/Devices/ExternalInput/Pic.cs
```diff
@@ -21,176 +23,288 @@
...
-    }
-
-    public Pic(Machine machine, bool initialized, Configuration configuration) : base(machine, configuration) {
-        _initialized = initialized;
+    public Pic(Machine machine, bool master) {
+        _master = master;
+        _machine = machine;
     }
 
     /// <summary>
```
maximilien-noal, line 57: obsolete documentation. It applied to the old code. Probably needs to be deleted.

src/Spice86.Core/Emulator/Devices/ExternalInput/Pic.cs
```diff
@@ -21,176 +23,288 @@
 /// <li>https://k.lse.epita.fr/internals/8259a_controller.html</li>
 /// </ul>
 /// </summary>
-public class Pic : DefaultIOPortHandler {
```
maximilien-noal, line 24: Dune exhibits the following behavior :
- music stops
- FMVs stop at the first frame
- Voice lines stops and then the game stops responding because no IRQs is being serviced at all around the 26 million instruction mark.
- Sometimes it doesn't exit the intro in order to start a new game after the Esc key is used.

PR #229, opened 2023-01-10
Feature/ems

maximilien-noal:
Code is from the Aeon Emulator.

Dune still runs fine.

How to run with EMS enabled, and enough memory to install the EMS Handler:
            "args": ["--Ems", "-k", "8192"],

### DIFF
--- a/src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
+++ b/src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
@@ -1,5 +1,6 @@
 ﻿namespace Spice86.Core.Emulator.Memory;
 
+using Spice86.Core.Emulator.Devices;
 using Spice86.Core.Emulator.InterruptHandlers;
 using Spice86.Core.Emulator.VM;
 
@@ -11,7 +12,7 @@ using System.Linq;
 /// <summary>
 /// Provides DOS applications with EMS memory.
 /// </summary>
-public sealed class ExpandedMemoryManager : InterruptHandler {
+public sealed class ExpandedMemoryManager : InterruptHandler, IDeviceCallbackProvider {
     /// <summary>
     /// Size of each EMS page in bytes.
     /// </summary>
@@ -71,6 +72,10 @@ public sealed class ExpandedMemoryManager : InterruptHandler {
     }
 
     public override byte Index => 0x67;
+
+    public bool IsHookable => throw new NotImplementedException();
+
+    public SegmentedAddress CallbackAddress { set => throw new NotImplementedException(); }
 
     public override void Run() {
         switch (_machine.Cpu.State.AH) {
@@ -773,5 +778,9 @@ public sealed class ExpandedMemoryManager : InterruptHandler {
         }
 
         return overlap ? (byte)0x92 : (byte)0;
+    }
+
+    public void InvokeCallback() {
+        throw new NotImplementedException();
     }
 }
--- a/src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
+++ b/src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
@@ -516,8 +516,6 @@ public sealed class ExpandedMemoryManager : InterruptHandler {
     /// Copies a block of memory.
     /// </summary>
     private void Move() {
-#warning make sure works with new paging system
-
         int length = (int)_machine.Memory.GetUint32(_machine.Cpu.State.DS, _machine.Cpu.State.SI);
 
         byte sourceType = _machine.Memory.GetByte(_machine.Cpu.State.DS, _machine.Cpu.State.SI + 4u);
--- a/src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
+++ b/src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
@@ -11,8 +11,7 @@ using System.Linq;
 /// <summary>
 /// Provides DOS applications with EMS memory.
 /// </summary>
-public sealed class ExpandedMemoryManager : InterruptHandler
-{
+public sealed class ExpandedMemoryManager : InterruptHandler {
     /// <summary>
     /// Size of each EMS page in bytes.
     /// </summary>
@@ -36,8 +35,7 @@ public sealed class ExpandedMemoryManager : InterruptHandler
     private readonly int[] mappedPages = new int[MaximumPhysicalPages] { -1, -1, -1, -1 };
     private readonly uint xmsBaseAddress;
 
-    public ExpandedMemoryManager(Machine machine) : base(machine)
-    {
+    public ExpandedMemoryManager(Machine machine) : base(machine) {
         this.pageOwners.AsSpan().Fill(-1);
         this._machine.Memory.SetString(0xF100, 0x000A, "EMMXXXX0");
         this._machine.Memory.Reserve(PageFrameSegment, PageSize * MaximumPhysicalPages);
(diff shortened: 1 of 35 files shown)


maximilien-noal:
A lot of games want XMS disabled, so EMS vs XMS being present should also be configurable.

maximilien-noal:
TODO:
- EmsFunctions -> Replace this with public function mapped in our Dictionary.
- Replace pointers with managed alternatives (Span, ref, ref struct, ...)
- XMS: Don't implement IMultiplexInterruptHandler from Aeon. Use our code that uses public and simple methods, so generated code can call it directly.
/// <summary>
    /// Defines an interrupt handler for a virtual machine on the 2F multiplex interrupt.
    /// </summary>
    public interface IMultiplexInterruptHandler : IVirtualDevice
    {
        /// <summary>
        /// Gets the interrupt handler's identifier selected by the value in the AH register.
        /// </summary>
        int Identifier { get; }

        /// <summary>
        /// Called when the interrupt handler should perform its action.
        /// </summary>
        void HandleInterrupt();
    }
    For example: int 21, ah 02 => DisplayOutput

src/Spice86.Core/Emulator/Memory/Memory.cs
```diff
@@ -1,27 +1,259 @@
...
         UInt32 = new(this);
     }
 
+    /// <summary>
+    /// Reserves a block of conventional memory.
+    /// </summary>
+    /// <param name="minimumSegment">Minimum segment of requested memory block.</param>
+    /// <param name="length">Size of memory block in bytes.</param>
+    /// <returns>Information about the reserved block of memory.</returns>
+    public ReservedBlock Reserve(ushort minimumSegment, uint length) {
```
kevinferrare, line 81: Not sure to understand why this is in Memory

src/Spice86.Core/Emulator/Memory/Memory.cs
```diff
@@ -1,27 +1,259 @@
...
+    }
+
+
+    /// <summary>
+    /// Writes a string to memory as a null-terminated ANSI byte array.
+    /// </summary>
+    /// <param name="segment">Segment to write string.</param>
+    /// <param name="offset">Offset to write string.</param>
+    /// <param name="value">String to write to the specified address.</param>
+    public void SetString(uint segment, uint offset, string value) => SetString(segment, offset, value, true);
```
kevinferrare, line 135: SetString / GetString are already available for memory based datastructures in MemoryBasedDataStructureWithBaseAddressProvider.GetZeroTerminatedString / SetZeroTerminatedString.

src/Spice86.Core/Emulator/Memory/Memory.cs
```diff
@@ -1,27 +1,259 @@
...
+                writePtr[2] = 0;
+                writePtr[3] = 0x90;
+                writePtr[4] = 0x90;
+                writePtr[5] = 0x90;
+
+                writePtr += 6;
+                length += 6;
+            }
+
+            writePtr[0] = 0x0F;
```
kevinferrare, line 116: Our CPU will not understand this. We use opcode FE38 which is the same dosbox uses.

src/Spice86.Core/Emulator/Memory/XmsMoveData.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace Spice86.Core.Emulator.Memory;
+
+using System.Runtime.InteropServices;
+
+/// <summary>
+/// In-memory structure with information about an XMS move request.
+/// </summary>
+[StructLayout(LayoutKind.Sequential, Pack = 1)]
```
kevinferrare, line 8: To convert to emulated memory structure? See DosMemoryControlBlock for example

src/Spice86.Core/Emulator/Memory/ExtendedMemoryManager.cs
```diff
@@ -0,0 +1,485 @@
...
+
+        IntPtr srcPtr = IntPtr.Zero;
+        IntPtr destPtr = IntPtr.Zero;
+
+        if (moveData.SourceHandle == 0) {
+            SegmentedAddress srcAddress = moveData.SourceAddress;
+            srcPtr = this._machine.Memory.GetPointer(srcAddress.Segment, srcAddress.Offset);
+        } else {
+            if (this.TryGetBlock(moveData.SourceHandle, out XmsBlock srcBlock)) {
+                srcPtr = this._machine.Memory.GetPointer((int)(XmsBaseAddress + (srcBlock).Offset + moveData.SourceOffset));
```
kevinferrare, line 435: I think code would be much simpler if instead of storing the block data in emulated memory (which cannot be accessed past 1M in real mode anyway) we stored them as byte array in a proper c# datastructure

src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
```diff
@@ -0,0 +1,777 @@
...
+
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Linq;
+
+/// <summary>
+/// Provides DOS applications with EMS memory.
+/// </summary>
+public sealed class ExpandedMemoryManager : InterruptHandler {
```
kevinferrare, line 14: Should probably be in Spice86.Core.Emulator.InterruptHandlers.Dos namespace

src/Spice86.Core/Emulator/Memory/Memory.cs
```diff
@@ -1,27 +1,259 @@
...
+    /// Video RAM mapping is technically up to 0xBFFF0 normally.
+    /// </remarks>
+    private const int VramUpperBound = 0xBFFF << 4;
+
+    /// <summary>
+    /// Size of conventional memory in bytes.
+    /// </summary>
+    public const uint ConvMemorySize = 1024 * 1024;
+
+    public bool EnableA20 { get; internal set; }
```
kevinferrare, line 36: This is actually an interesting property. 
 - If disabled, memory accesses should wrap when address > 1M
 - If enabled, memory addresses > 1M give access to a bit more ram (HMA, more info here: http://flint.cs.yale.edu/feng/cos/resources/BIOS/himem_A20.htm)

src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
```diff
@@ -0,0 +1,777 @@
...
+            if (this.mappedPages[i] != -1) {
+                Span<byte> src = this.GetLogicalPage(this.mappedPages[i]);
+                Span<byte> dest = this.GetMappedPage(i);
+                src.CopyTo(dest);
+            }
+        }
+    }
+
+    private Span<byte> GetMappedPage(int physicalPageIndex) => this._machine.Memory.Span.Slice((PageFrameSegment << 4) + (physicalPageIndex * PageSize), PageSize);
+    private Span<byte> GetLogicalPage(int logicalPageIndex) => this._machine.Memory.Span.Slice((int)this.xmsBaseAddress + (logicalPageIndex * PageSize), PageSize);
```
kevinferrare, line 591: Code would be way simpler overall if ems / xms pages are stored outside of memory

kevinferrare (commented):
I think aeon code is complex because it is storing ems / xms extra memory in the emulated ram.
Looking at how it's done elsewhere I am not sure this is necessary and not doing so could make code much simpler.

src/Spice86.Core/Emulator/Memory/XmsMoveData.cs
maximilien-noal: Probably. It's also very close to what DOSBox does:

    #pragma pack (1)
    #endif
    struct XMS_MemMove{
    	uint32_t length;
    	uint16_t src_handle;
    	union {
    		RealPt realpt;
    		uint32_t offset;
    	} src;
    	uint16_t dest_handle;
    	union {
    		RealPt realpt;
    		uint32_t offset;
    	} dest;
    
    } GCC_ATTRIBUTE(packed);
    #ifdef _MSC_VER
    #pragma pack ()
    #endif

src/Spice86.Core/Emulator/Memory/ExpandedMemoryManager.cs
maximilien-noal: Done !

maximilien-noal:
PR closed. Future PR will only include EMS.

src/Spice86.Core/Emulator/Memory/Memory.cs
maximilien-noal: Fixed in feature/xms

PR #235, opened 2023-01-13
Emulation: Support for Alpha Waves / Continuum

### LINKED ISSUE #233: Alpha Waves / Continuum does not run.
**Describe the bug**

Alpha Waves / Continuum does not run.

It should:
https://www.mobygames.com/game/continuum/techinfo
Minimum CPU Class Required | Intel 8088 / 8086
Minimum OS Class Required | PC/MS-DOS 2.0
Minimum RAM Required | 512 KB
Media Type | 3.5" Floppy Disk,  5.25" Floppy Disk
Video Modes Supported | CGA, EGA, Hercules, Tandy / PCjr, VGA
Sound Devices Supported | Adlib, PC Speaker, Tandy / PCjr
Input Devices Supported | Keyboard
Controller Types Supported | Analog Joystick
Number of Offline Players | 1-2 Players

**To Reproduce**

Try to run it.

**Expected behavior**

maximilien-noal:
fixes #233 (eventually)


LowLevelMahn:
`Ega2bpp640x400 = 0x0D,`

another thing: isn't it 4bpp ?

maximilien-noal:
Thanks, I corrected it.

Guess I'll have to extend 640x200 to 640x400 on the screen.

maximilien-noal:
TODO:
- Set default EGA palette in EGA mode (see VgaDac.cs)
- Implement VRAM, and VRAM copy to main memory
- Pass the videoMode to the VideoBufferViewModel, and introduce specific
  Presenters to the UI.
- Try to scale 640x200 to it's real aspect ratio
- Fix unhandled IO port 980 (0x03D4). Exception happens on CPu Cycle
  number 604.

LowLevelMahn:
here is a small Watcom/C example for programming the EGA 0dh mode by "Lameguy" (http://lameguy64.net/?page=dosstuff)
maybe a good AND still small enough testcase

[EGATEST.zip](https://github.com/OpenRakis/Spice86/files/10416896/EGATEST.zip)

compiles out of the box with OpenWatcom V2: 
https://github.com/open-watcom/open-watcom-v2/releases/tag/Current-build
https://github.com/open-watcom/open-watcom-v2/releases/download/Current-build/open-watcom-2_0-c-win-x64.exe (just unpack the exe with 7zip or something else - no installation needed) 

see EGATEST.zip/build.bat for building

maximilien-noal:
I was thinking of searching the same kind of thing. This will help a lot !

LowLevelMahn:
> I was thinking of searching the same kind of thing. This will help a lot !

took me a while to find all that stuff - EGA isn't that present on the internet - only VGA is big

maximilien-noal:
@LowLevelMahn This is on standby since support for Betrayal at Krondor touches the same files (for example, it also needs EGA (namely the EGA Sequencer)).

LowLevelMahn:
> @LowLevelMahn This is on standby since support for Betrayal at Krondor touches the same files (for example, it also needs EGA (namely the EGA Sequencer)).

its ok - im so busy with many other things - reversing is currently only a 30min per weeks thing ... but great that you work on Krondor

maximilien-noal:
So, resuming this. There is only one writeByte that is problematic at the moment.

maximilien-noal:
PR ready for review.

@LowLevelMahn I'll make another PR in order to fix the remaining color glitches.

src/Spice86.Aeon/Emulator/Video/Ports.cs
```diff
@@ -20,6 +20,8 @@ public static class Ports
         public const int MiscOutputRead = 0x03CC;
         public const int GraphicsControllerAddress = 0x03CE;
         public const int GraphicsControllerData = 0x03CF;
+        public const int CrtControllerAddressAltMirror1 = 0x03D0;
+        public const int CrtControllerAddressAltMirror2 = 0x03D2;
         public const int CrtControllerAddressAlt = 0x03D4;
         public const int CrtControllerDataAlt = 0x03D5;
```
JorisVanEijden, line 26: You may want to add 0x03D1 and 0x03D3, that mirror 0x03D5 as well.

src/Spice86.Core/Emulator/Devices/Video/AeonCard.cs
```diff
@@ -254,10 +261,10 @@ public class AeonCard : DefaultIOPortHandler, IVideoCard, IAeonVgaCard, IDisposa
                 if (previousVerticalEnd != CrtController.VerticalDisplayEnd) {
                     ChangeVerticalEnd();
                 }
-
                 break;
             default:
-                throw new InvalidOperationException($"Writing 0x{value:X2} to port 0x{port:X4} is not (yet) supported by this VGA controller.");
+                base.WriteByte(port, value);
```
JorisVanEijden, line 266: I'm not sure. 
There are 2 cases in which we hit this situation:
1. the `AeonCard.WriteByte` method is being called with a port it did not register itself for.
2. the `AeonCard.WriteByte` method is being called with a port it can only read from.

In case 1 I think we should still have an exception. For case 2 calling the base class seems fine.

This also loses logging of the value being written. Maybe that should be included in the base class?

maximilien-noal: For case 1, the base class handles this. If there is no-one registered, and -f was passed on the command line, an exception is thrown.
For case 2, the base class handles this, if -f was passed on the command line, and the child class called the base class.

JorisVanEijden: Case 1 is when the Aeon class did _not_ register the port, but the **method** in the Aeon class is called with that port anyway. 
Are we sure we want to ignore this if the -f flag is not set?

maximilien-noal: Ah, I see. Called by overrides. Good point.

I think this somehow should be handled by the Base Class for every case, not just the Aeon Card. Maybe we'll have to call an inherited guard method in all child class... Doesn't feel like the cleanest solution.

JorisVanEijden: I think the most likely cause would be the Word and DWord versions of the port accessing API.
We can fix that by changing the `IOPortDispatcher` to get he port for each byte and only call WriteByte.

Not something to fix in this PR probably.

LowLevelMahn:
> No remaining unhandled ports, and the game is "playable" (there's still color problems).
> 
> Edit: @LowLevelMahn I think I must compare with DOSBox in order to fix the remaining color glitches.

Alpha Waves uses the EGA mode 320x200x16 combinded with VGA palett color changes to brigthen the colors - maybe its the combination

JorisVanEijden (approved):

PR #246, opened 2023-01-20
Feature/xms

maximilien-noal:


maximilien-noal:
Dune now works with XMS mode. But only uses the Bios service to copy extended memory.
Nothing in the Xms class itself, not even the callback.

Tested with: "DNCDPRG.EXE XMS ADP220 SBP2227"

PR #248, opened 2023-02-09
Just enough VGA and BIOS to get Krondor to the not-enough-memory exit

JorisVanEijden:
- Introduce BIOS memory as a separate class (segment 0x40)
- Introducs VGA Bios to VgaCard
- Add 3 standard fonts to VGA BIOS (lazy loaded)
- Fix negative dos files index

This lays a foundation for further bios & vga additions

PR #249, opened 2023-02-09
Dynamic logging level

maximilien-noal:
Signed-off-by: Maximilien Noal <noal.maximilien@gmail.com>

Also, they can be silenced now (for better performance), and controlled by the UI.


maximilien-noal:
Lower priority compared to:
https://github.com/OpenRakis/Spice86/pull/248

(because this one will probably create merge conflicts if merged first)

maximilien-noal:
The better performance (when they are silenced) helps with debugging Lost Eden (a good use case for feature/ems)

PR #255, opened 2023-02-13
CPU: ENTER instruction

### LINKED ISSUE #250: CPU: Enter instruction
This is both a 16 and 32 bit instruction.

https://github.com/bochs-emu/Bochs/blob/17e4e25e2f3a65c2fe0c39ba24e5bfd857f4eba8/bochs/cpu/stack16.cc#L178

https://c9x.me/x86/html/file_module_x86_id_78.html

> Creates a stack frame for a procedure. The first operand (size operand) specifies the size of the stack frame (that is, the number of bytes of dynamic storage allocated on the stack for the procedure).
>
>The second operand (nesting level operand) gives the lexical nesting level (0 to 31) of the procedure. The nesting level determines the number of stack frame pointers that are copied into the "display area" of the new stack frame from the preceding frame. Both of these operands are immediate values.
>
>The stack-size attribute determines whether the BP (16 bits) or EBP (32 bits) register specifies the current frame pointer and whether SP (16 bits) or ESP (32 bits) specifies the stack pointer.
>
>The ENTER and companion LEAVE instructions are provided to support block structured languages. The ENTER instruction (when used) is typically the first instruction in a procedure and is used to set up a new stack frame for a procedure. The LEAVE instruction is then used at the end of the procedure (just before the RET instruction) to release the stack frame.
>
>If the nesting level is 0, the processor pushes the frame pointer from the EBP register onto the stack, copies the current stack pointer from the ESP register into the EBP register, and loads the ESP register with the current stack-pointer value minus the value in the size operand. For nesting levels of 1 or greater, the processor pushes additional frame pointers on the stack before adjusting the stack pointer. These additional frame pointers provide the called procedure with access points to other nested frames on the stack. See "Procedure Calls for Block-Structured Languages" in Chapter 6 of the IA-32 Intel Architecture Software Developer's Manual, Volume 1, for more information about the actions of the ENTER instruction.

maximilien-noal:
Signed-off-by: Maximilien Noal <noal.maximilien@gmail.com>

Closes #250 

Fixes Lands Of Lore and Stunts 4D. Although, they don't run for other reasons (INT2F and VGA services not implemented, respectively)

### DIFF
--- a/src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions16.cs
+++ b/src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions16.cs
@@ -535,17 +535,17 @@ public class Instructions16 : Instructions16Or32 {
         ushort sp = State.SP;
         ushort bp = State.BP;
         sp -= 2;
-        Memory.SetUint16(MemoryUtils.ToPhysicalAddress((ushort)(State.SS + sp), 0), State.BP);
+        Memory.UInt16[MemoryUtils.ToPhysicalAddress((ushort)(State.SS + sp), 0)] = State.BP;
         State.BP = (ushort)(State.SP - 2);
         if (level > 0) {
             // do level-1 times
             for (int i = 1; i < level; i++) {
                 sp -= 2;
                 bp -= 2;
-                Memory.SetUint16(MemoryUtils.ToPhysicalAddress((ushort)(State.SS + sp), 0), Memory.GetUint16(MemoryUtils.ToPhysicalAddress((ushort)(State.SS + bp), 0)));
+                Memory.UInt16[MemoryUtils.ToPhysicalAddress((ushort)(State.SS + sp), 0)] = Memory.UInt16[MemoryUtils.ToPhysicalAddress((ushort)(State.SS + bp), 0)];
             }
             sp -= 2;
-            Memory.SetUint16(MemoryUtils.ToPhysicalAddress((ushort)(State.SS + sp), 0), Memory.GetUint16(MemoryUtils.ToPhysicalAddress((ushort)(State.BP), 0)));
+            Memory.UInt16[MemoryUtils.ToPhysicalAddress((ushort)(State.SS + sp), 0)] = Memory.UInt16[MemoryUtils.ToPhysicalAddress(State.BP, 0)];
         }
         sp -= bytes;
         State.ESP = State.ESP | sp;
--- a/src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions16.cs
+++ b/src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions16.cs
@@ -529,27 +529,39 @@ public class Instructions16 : Instructions16Or32 {
     }
 
     public override void Enter() {
-        ushort bytes = Cpu.NextUint16();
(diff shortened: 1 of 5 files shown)


src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions16.cs
```diff
@@ -526,6 +528,30 @@ public class Instructions16 : Instructions16Or32 {
...
     }
 
+    public override void Enter() {
+        ushort bytes = Cpu.NextUint16();
+        byte level = Cpu.NextUint8();
+        level &= 0x1F;
+        ushort sp = State.SP;
+        ushort bp = State.BP;
+        sp -= 2;
+        Memory.UInt16[MemoryUtils.ToPhysicalAddress((ushort)(State.SS + sp), 0)] = State.BP;
```
JorisVanEijden, line 538: This looks strange.
Did you mean:
```suggestion
        Memory.UInt16[(ushort)(State.SS + sp), 0] = State.BP;
```
?

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt2fHandler.cs
```diff
@@ -30,14 +30,13 @@ public class DosInt2fHandler : InterruptHandler {
...
-        _dispatchTable.Add(0x43, new Callback(0x43, NoOp));
+        _dispatchTable.Add(0x43, new Callback(0x43, GetSetFileAttributes));
     }
 
     /// <summary>
-    /// This is an INT2F function that even DOSBox doesn't implement.
     /// Right now, a NOP is sufficient in order to make some games (eg. Dune 2) work.
     /// </summary>
-    public void NoOp() {
+    public void GetSetFileAttributes() {
```
JorisVanEijden, line 40: This could also use a `TODO` comment. 
Makes it easier to find places to work on.

maximilien-noal: Done.

src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions16.cs
maximilien-noal: Good catch. This is fixed.

src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions16.cs
```diff
@@ -526,6 +528,42 @@ public class Instructions16 : Instructions16Or32 {
         Cpu.Out16(State.DX, State.AX);
     }
 
+    public override void Enter() {
```
kevinferrare, line 531: I think this is not correct because stackSize is 16 for 16bits instructions and 32 for 32bit.
For the 32bits I would rather use the algorithm descried here: https://xem.github.io/minix86/manual/intel-x86-and-64-manual-vol1/o_7281d5ea06a5b67a-159.html
And adapt it with BP instead of EBP and SP instead of ESP for 16 bits.

maximilien-noal: Done.

kevinferrare (approved):

PR #257, opened 2023-02-14
Allow specifying a custom interrupt handler location

JorisVanEijden:
Allow device drivers to specify a custom memory location for their interrupt handler stub.

This is useful because by convention the Device Driver Header is installed at offset 0 of the same segment. 
For instance if a program wants to know if an EMS driver is installed it can call INT 21 35 to get the location of the INT 67 handler and then check offset 0xA of the returned segment to get the name of the driver.

See: https://www.lo-tech.co.uk/wiki/LIM_Expanded_Memory_Specification_V4:_Appendix_B


maximilien-noal:
Merged also, nice !

Issue #228, opened 2023-01-08
Cpu: 32 bit unit tests from qemu 


maximilien-noal:

JorisVanEijden:
Are there any plans in motion for this?

maximilien-noal:
This issue is not being worked on, feel free to grab it !

Also this issue should be renamed ' take the test executable from the qemu project and test 32 bit support with it', as I understand it.

@kevinferrare knows more about this.

JorisVanEijden:
How about using this? https://github.com/barotto/test386.asm

It appears to be a test executable specifically written for testing x86 emulators.

kevinferrare:
I considered it at one point but it is testing protected mode which we don't support. We could edit the asm to remove that though.
Otherwise qemu has nice tests that you can recompile with only the features you want: https://github.com/JonathanSalwan/Triton/blob/master/src/testers/unittests/misc/qemu/test-i386.c

maximilien-noal:
https://github.com/TomHarte/ProcessorTests (contains the 8088 tests, too, the only part, apart from the approach, that might be revelant)

maximilien-noal:
> Are there any plans in motion for this?

There is some. This PR has been merged: https://github.com/OpenRakis/Spice86/pull/1049

We use a modified version of test386.asm now, without protected mode tests.

maximilien-noal:
Interesting repo real mode tests of the 286 CPU: https://github.com/SingleStepTests/80286

maximilien-noal:
Tests for the CPU are being worked on by @kevinferrare and @codengine

PR #260, opened 2023-02-18
Imul instruction addition

JorisVanEijden:
Added the 0x0F 0xAF variation of the IMUL instruction.

No tests yet, will be done in #228

### DIFF
--- a/src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions32.cs
+++ b/src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions32.cs
@@ -222,7 +222,7 @@ public class Instructions32 : Instructions16Or32 {
     public override void ImulRmReg16Or32() {
         // IMUL32 r32 rm32
         ModRM.Read();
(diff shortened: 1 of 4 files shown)


src/Spice86.Core/Emulator/CPU/InstructionsImpl/Instructions32.cs
```diff
@@ -218,6 +218,12 @@ public class Instructions32 : Instructions16Or32 {
         long result = Alu.Imul32(value, (int)ModRM.GetRm32());
         ModRM.R32 = (uint)result;
     }
+    
+    public override void ImulRmReg16Or32() {
+        // IMUL32 r32 rm32
+        ModRM.Read();
+        ImulRmVal((short)ModRM.R32);
```
kevinferrare, line 225: Shouldnt it be (int) ?

JorisVanEijden: Fixed

PR #265, opened 2023-02-20
Initial prefix 67 support

JorisVanEijden:
This introduces opcode prefix 67 (address size) support, including the decoding of the SIB byte.

This supersedes PR #262 

To keep the PR small I have not yet identified and adjusted all the places where this should be used. I want to get a stable implementation base in first.

### DIFF
--- a/src/Spice86.Core/Emulator/CPU/CPU.cs
+++ b/src/Spice86.Core/Emulator/CPU/CPU.cs
@@ -41,19 +41,11 @@ public class Cpu {
     private readonly Instructions16 _instructions16;
     private readonly Instructions32 _instructions32;
     private Instructions16Or32 _instructions16Or32;
-    private int _addressSize;
 
     /// <summary>
     /// Address size for the currently executing instruction.
     /// </summary>
-    public int AddressSize {
-        get => _addressSize;
-        private set {
-            if (value != 16 && value != 32)
-                throw new ArgumentOutOfRangeException(nameof(value), "Address size must be 16 or 32");
-            _addressSize = value;
-        }
-    }
+    public int AddressSize { get; private set; }
 
     public CallbackHandler? CallbackHandler { get; set; }
 
--- a/src/Spice86.Core/Emulator/CPU/CPU.cs
+++ b/src/Spice86.Core/Emulator/CPU/CPU.cs
@@ -41,6 +41,20 @@ public class Cpu {
     private readonly Instructions16 _instructions16;
     private readonly Instructions32 _instructions32;
     private Instructions16Or32 _instructions16Or32;
+    private int _addressSize;
+
+    /// <summary>
+    /// Address size for the currently executing instruction.
+    /// </summary>
+    public int AddressSize {
+        get => _addressSize;
+        private set {
+            if (value != 16 && value != 32)
+                throw new ArgumentOutOfRangeException(nameof(value), "Address size must be 16 or 32");
+            _addressSize = value;
+        }
+    }
+
     public CallbackHandler? CallbackHandler { get; set; }
 
     // When true will crash if an interrupt targets code at 0000:0000
@@ -74,6 +88,7 @@ public class Cpu {
         _instructions16 = new Instructions16(machine, Alu, this, _memory, _modRM, StaticAddressesRecorder);
         _instructions32 = new Instructions32(machine, Alu, this, _memory, _modRM, StaticAddressesRecorder);
         _instructions16Or32 = _instructions16;
+        AddressSize = 16;
     }
 
     public void ExecuteNextInstruction() {
@@ -92,8 +107,9 @@ public class Cpu {
                 HandleCpuException(e);
             }
         }
-        // Reset to 16 bit operand size
+        // Reset to 16 bit operand and address size
         _instructions16Or32 = _instructions16;
+        AddressSize = 16;
         State.ClearPrefixes();
         StaticAddressesRecorder.Commit();
         State.IncCycles();
@@ -231,7 +247,7 @@ public class Cpu {
             case 0x8D:
             case 0x8E:
             case 0x8F:
-                Jcc(TestJumpCondition(subcode));
+                Jcc(TestJumpCondition(subcode), AddressSize);
                 break;
             case 0x90:
             case 0x91:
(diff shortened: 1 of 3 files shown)


src/Spice86.Core/Emulator/CPU/CPU.cs
```diff
@@ -40,6 +40,20 @@ public class Cpu {
...
     private Instructions16Or32 _instructions16Or32;
+    private int _addressSize;
+
+    /// <summary>
+    /// Address size for the currently executing instruction.
+    /// </summary>
+    public int AddressSize {
+        get => _addressSize;
+        private set {
+            if (value != 16 && value != 32)
```
kevinferrare, line 51: How about using an enum to avoid the if penalty at each instruction?

src/Spice86.Core/Emulator/CPU/ModRM.cs
```diff
@@ -160,4 +170,59 @@ public class ModRM {
...
+        return (uint)(@base + scale * index);
+    }
+
+    private int ComputeSibIndex(int indexRegister) {
+        uint result = indexRegister switch {
+            0 => _state.EAX,
+            1 => _state.ECX,
+            2 => _state.EDX,
+            3 => _state.EBX,
+            4 => 0,
```
kevinferrare, line 205: Isn't Index 4 is invalid?
http://www.c-jump.com/CIS77/CPU/x86/X77_0100_sib_byte_layout.htm

src/Spice86.Core/Emulator/CPU/ModRM.cs
```diff
@@ -93,14 +93,24 @@ public class ModRM {
...
+                _ => 0
+            };
+            MemoryOffset = (ushort?)(ComputeOffset16(bpForRm6) + disp);
+        } else {
+            int disp = mode switch {
+                1 => (sbyte)_cpu.NextUint8(),
+                2 => (int)_cpu.NextUint32(),
+                _ => 0
+            };
+            MemoryOffset = (ushort?)(ComputeOffset32(mode, _registerMemoryIndex) + disp);
```
kevinferrare, line 111: ComputeOffset32 potentially needs the SIB, so I don't think it can work like that, it needs to be called before disp computation isnt it?

src/Spice86.Core/Emulator/CPU/ModRM.cs
```diff
@@ -160,4 +170,59 @@ public class ModRM {
             _ => throw new InvalidModeException(_machine, _registerMemoryIndex)
         };
     }
+
+    private int ComputeOffset32(int mode, int rm) {
```
kevinferrare, line 174: Are you sure this is supposed to be signed?

src/Spice86.Core/Emulator/CPU/ModRM.cs
kevinferrare, line 111: I think you should check that the uint result is no more than FFFF, and raise an int if that's the case (intel spec says so somehow)

src/Spice86.Core/Emulator/CPU/ModRM.cs
JorisVanEijden: No. It means "no index register is used". Setting it to 0 amounts to the same result as not using a register, since the formula is always base + (index * scale).

src/Spice86.Core/Emulator/CPU/ModRM.cs
JorisVanEijden: Yes you are correct, the displacement bytes are _after_ the SIB byte :(
I think we can fix that by switching the order.

JorisVanEijden: Yes I agree. I have another branch I'm working on that implements exception handling and interrupt throwing.
I will try to find as many places as I can to then add throwing the appropriate exceptions.

src/Spice86.Core/Emulator/CPU/CPU.cs
JorisVanEijden: Hm, I missed that it is setting it to 16 for every instruction.
That whole check can go as far as I'm concerned.

src/Spice86.Core/Emulator/CPU/ModRM.cs
JorisVanEijden: No. I'll change it. I can't really find it in the manual but it makes no sense for it to be negative since we cast it to ushort before using it anyway.

src/Spice86.Core/Emulator/CPU/CPU.cs
```diff
@@ -1266,8 +1286,14 @@ private static bool IsStringOpUpdatingFlags(int stringOpCode)
     /// <summary>
     /// Jumps handling
     /// </summary>
-    private void Jcc(bool jump) {
-        sbyte address = (sbyte)NextUint8();
+    private void Jcc(bool jump, int offsetSize = 8) {
```
maximilien-noal, line 1289: Shouldn't this be public so low-level C# overrides of ASM code can call it ?

src/Spice86.Core/Emulator/CPU/CPU.cs
```diff
@@ -1219,8 +1235,12 @@ private static bool IsStringOpUpdatingFlags(int stringOpCode)
     }
 
     private bool TestJumpConditionCXZ() {
```
maximilien-noal, line 1237: Shouldn't this be public so low-level C# overrides of ASM code can call it ?

src/Spice86.Core/Emulator/CPU/CPU.cs
```diff
@@ -1345,6 +1371,9 @@ private static bool IsStringOpUpdatingFlags(int stringOpCode)
             case 0x66:
                 _instructions16Or32 = _instructions32;
                 break;
+            case 0x67:
+                AddressSize = 32;
```
maximilien-noal, line 1375: Shouldn't this be done in a public method so low-level C# overrides of ASM code can call this instruction directly ?

maximilien-noal (approved):
Some methods to be made public I think. Otherwise, seems good to me !

src/Spice86.Core/Emulator/CPU/CPU.cs
kevinferrare: For generated code, you will know whether you are in 16 or 32 bits beforehand so you will either do CX ==0 or ECX == 0 but no if

src/Spice86.Core/Emulator/CPU/ModRM.cs
```diff
@@ -93,15 +93,31 @@ public class ModRM {
...
+        if (_cpu.AddressSize == 16) {
+            short displacement = mode switch {
+                1 => (sbyte)_cpu.NextUint8(),
+                2 => (short)_cpu.NextUint16(),
+                _ => 0
+            };
+            ushort offset = ComputeOffset16(mode);
+            MemoryOffset = (ushort)(offset + displacement);
+        } else {
+            long offset = ComputeOffset32(mode, _registerMemoryIndex);
```
kevinferrare, line 106: Why long?

JorisVanEijden: Because that saves from having to cast the result of adding the displacement back to uint (for the comparison) and then back to ushort.
I did miss a check to see if the offset becomes negative. Which I;ve now added. As well as updated the exception.

PR #285, opened 2023-03-21
fix/color palette window

maximilien-noal:
- VGA DAC uses an Array of RGBs instead of an UnsafeBuffer.
- It also uses a Span instead of a pointer.
- VideoMode class also uses a ReadOnlySpan<Rgb> for its Palette property
- Rgb class defines the static implicit operator to uint, so Presenters still work.
- UI: Color Palette window works again.

Tested with:
- Dune

Performance: no regression to my knowledge.


JorisVanEijden:
I'm not sure I agree with the approach here.
You convert incoming 6-bit colors to 8-bit. (You forgot to convert outgoing colors from 8-bit to 6-bit again.)
All so that you can store the colors in a way that's easier on the debug.

I'd rather see an approach that reads, stores and writes the original 6-bit colors and let the debug system convert them to 8-bit when it needs them (either through calculation or lookup table).

You could also keep 2 stores. A `uint` one for reading and writing and the presenter, and an `Rgb` one for the debug display.
Or even easier, create a `fromUint` extension method for `Color` and use that in the PaletteViewModel for `fill.Color = Color.FromRgb`

Or am I missing something?

maximilien-noal:
All very good points, I'll implement them. 

Eventually, I'd like user overrides that call the video handler directly to use RGB instead of Uint, it would be easier for anyone to read and write.

But this is not the aim of this PR. 

This PR aimed at three things :

Remove unsafe code in the Dac. 
Make the color palette UI work again. 
Reintroduce the shared Rgb class. 

Color already has a fromuint method.

src/Spice86.Shared/Rgb.cs
```diff
@@ -25,4 +25,12 @@ public class Rgb {
     public override string ToString() {
         return System.Text.Json.JsonSerializer.Serialize(this);
     }
+
+    public static implicit operator uint(Rgb v) {
+        return ToUint(v);
+    }
+
+    public static uint ToUint(Rgb v) {
+        return (uint)(v.G << 16 | v.B << 8 | v.R);
```
JorisVanEijden, line 34:
```suggestion
        return (uint)(v.R << 16 | v.G << 8 | v.B);
```

src/Spice86.Aeon/Emulator/Video/Dac.cs
```diff
@@ -70,70 +64,66 @@ public byte WriteIndex
...
-                return (byte)((color >> 2) & 0x3F);
+                case 1:
+                    return color.G;
+                case 2:
+                    return color.B;
             }
+
+            _readChannel = 0;
+            _readIndex++;
+            return color.R;
```
JorisVanEijden, line 79:
```suggestion
                    return color.R;
                case 2:
                    return color.G;
            }

            _readChannel = 0;
            _readIndex++;
            return color.B;
```
_readChannel is updated before being used, so 1 is the first color (R)

src/Spice86.Aeon/Emulator/Video/Dac.cs
```diff
@@ -70,70 +64,66 @@ public byte WriteIndex
...
+                // We could shift by 2 instead, but while it's faster,
+                // it may not be as accurate.
+                case 1:
+                    color.G = _lightCorrectedValuesLookupTable[value];
+                    break;
+                case 2:
+                    color.B = _lightCorrectedValuesLookupTable[value];
+                    break;
+                default:
+                    color.R = _lightCorrectedValuesLookupTable[value];
```
JorisVanEijden, line 101:
```suggestion
                    color.R = _lightCorrectedValuesLookupTable[value];
                    break;
                case 2:
                    color.G = _lightCorrectedValuesLookupTable[value];
                    break;
                default:
                    color.B = _lightCorrectedValuesLookupTable[value];
```
_writeChannel is increased before being used, so 1 is the first color (R)

JorisVanEijden (changes requested):

maximilien-noal:
Requested changes are implemented. Thanks for your feedback. :)

I didn't remove the RGB class after all from the DAC, as Color.FromUint32 expects a different layout, and an alpha channel.

Latest changes on top:

- The Read method from the DAC class is the very same as the master branch (after all, RGB can be converted to an uint value)
- I removed the lookup table in the DAC.

JorisVanEijden:
dac.Write(0b01010100) should store 0b0010100, yet when I dac.Read() it I get 0b0010101.

Perhaps you should split the 6-bit (from io ports) and 8-bit (from overrides and debug) reading and writing into separate methods. Both accessing the same internal 6-bit storage.

maximilien-noal:
Thanks, I'll look into it and add unit tests.

For performance, I'd rather store it at 8 bit. Rendering reads the palette a lot of time, and not needing to convert to from 6 to 8 bit is a plus.

src/Spice86.Aeon/Emulator/Video/Dac.cs
```diff
@@ -1,139 +1,122 @@
...
 
         /// <summary>
         /// Reads the next channel in the current color.
         /// </summary>
         /// <returns>Red, green, or blue channel value.</returns>
-        public byte Read()
-        {
-            unsafe
+        public byte Read() {
+            uint color = _palette[_readIndex];
```
kevinferrare, line 60: Why not using the rgb class and its accessors here?

maximilien-noal: Without unit tests, it was a way to have a minimal set of changes, and therefore no regressions.

That didn't work, so I can use the accessors and Rgb class again.

JorisVanEijden:
> For performance, I'd rather store it at 8 bit. Rendering reads the palette a lot of time, and not needing to convert to from 6 to 8 bit is a plus.

Sure, you could keep a 6bit and an 8bit store. And do conversions only on write.

src/Spice86.Shared/Rgb.cs
```diff
@@ -25,4 +25,12 @@ public class Rgb {
     public override string ToString() {
         return System.Text.Json.JsonSerializer.Serialize(this);
     }
+
+    public static implicit operator uint(Rgb v) {
+        return ToUint(v);
+    }
+
+    public static uint ToUint(Rgb v) {
+        return (uint)(v.R << 16 | v.G << 8 | v.B);
```
kevinferrare, line 34: Maybe create a toRgb() method in the class and call it from here and from toArgb() ?

src/Spice86.Aeon/Emulator/Video/Dac.cs
maximilien-noal: Done.

src/Spice86.Aeon/Emulator/Video/Dac.cs
```diff
@@ -1,139 +1,122 @@
...
+            _readChannel++;
```
kevinferrare, line 62: I feel it would be more intuitive to increase the read channel at the end and to have a switch between (0, 1, 2), like arrays indexes

src/Spice86.Shared/Rgb.cs
maximilien-noal: Done.

src/Spice86.Aeon/Emulator/Video/Dac.cs
maximilien-noal: Done.

JorisVanEijden:
Here's a simple test:
```
 [Fact]
    public void TestReadingDacReturns6BitData() {
        // Arrange
        var dac = new Dac();
        
        // Act
        for (byte i = 0; i < byte.MaxValue; i++) {
            dac.Write(i);
        }
        
        // Assert
        for (byte i = 0; i < byte.MaxValue; i++) {
            byte expected = (byte)(i & 0b00111111);
            byte result = dac.Read();
            result.Should().Be(expected, $"The same 6 low bits of 0x{i:X2} that were written should be read back");
        }
    }

File: src/Spice86.Shared/Rgb.cs [hunk 7]
```diff
@@ -4,25 +4,78 @@
...
-    }
+    /// <summary>
+    /// Blue channel, on 6 bit
+    /// </summary>
+    public byte B6 { get; set; }
 
-    public uint ToArgb() {
-        return 0xFF000000 | (uint)R << 16 | (uint)G << 8 | B;
-    }
+    public uint ToRgba() => 0x000000FF | ToRgb();
```
JorisVanEijden, line 38: I think the ToRgb needs to be shifted left 8 bits.
Or `R8 << 24 | G8 << 16 | B8 << 8 | 0xFF`

src/Spice86.Shared/Rgb.cs
```diff
@@ -4,25 +4,78 @@
...
+    
+
+    public uint ToBgra() => (uint)(B8 << 16 | G8 << 8 | R8) | 0x000000FF;
```
JorisVanEijden, line 41:
```suggestion
    public uint ToBgra() => (uint)(B8 << 24 | G8 << 16 | R8 << 8 | 0xFF);
```

src/Spice86.Shared/Rgb.cs
maximilien-noal: Change applied. Thanks !

maximilien-noal:
PR entirely reworked so only the UI (and future user code) can use RGBs.

In other words, the DAC is the same as in the master branch, same for Presenters.

We only define an extension method for uint, used by the UI (also available to user code), in order to convert them to RGB.

The RGB class is a readonly record struct.

PR #286, opened 2023-03-26
Video memory access crash prevention

JorisVanEijden:
Make sure we have the right bitmap, memory and presenter. This should avoid a number of illegal memory access crashes.


maximilien-noal:
Merged ! :)

PR #291, opened 2023-04-01
Fix ega colors

JorisVanEijden:
The 4-bit depth presenter was looking up colors in the vga palette, using the actual 6-bit ega color representations as index.

While this may work with an unchanged palette, when a program (such as the attached demo) changes the ega pallette, the resulting colors no longer match.

This fix uses the formula from https://moddingwiki.shikadi.net/wiki/EGA_Palette to convert the 6-bit ega color format to 24-bit rgb.

[L43-1.zip](https://github.com/OpenRakis/Spice86/files/11128539/L43-1.zip)


JorisVanEijden:
Back to draft. This fixes the attached demo, but breaks krondor even worse :(

PR #293, opened 2023-04-04
Light DOS refactoring, and some more INT2F, Mouse, and BIOS services

maximilien-noal:


src/Spice86.Core/Emulator/Memory/Memory.cs
```diff
@@ -252,6 +252,11 @@ public class Memory {
         get;
     }
 
+    /// <summary>
+    /// Is A20 gate rollover enabled or not
+    /// </summary>
+    public bool IsA20Enabled { get; set; }
```
kevinferrare, line 258: Depending on that, physical addresses should wrap around the ram. Not sure this is something to do in this PR.
With it disabled memory address should wrap with a modulo of 0xFFFFF
With it enabled, the modulo is (0xFFFF*16 + 0xFFFF) which is 0x10FFEF

src/Spice86.Core/Emulator/InterruptHandlers/Bios/SystemBiosInt15Handler.cs
```diff
@@ -20,14 +23,25 @@ public class SystemBiosInt15Handler : InterruptHandler {
...
+    /// Reports extended memory size in AX.
     /// </summary>
     public void GetExtendedMemorySize() {
-        //We've got no extended memory (yet)
         _state.AX = 0;
     }
 
+    public void CopyExtendedMemory(bool calledFromVm) {
+        bool enabled = _memory.IsA20Enabled;
+        _machine.Memory.EnableOrDisableA20Gate(true);
```
kevinferrare, line 34: why not _memory.IsA20Enabled = true (seems to do the same thing)?

src/Spice86.Core/Emulator/InterruptHandlers/Bios/SystemBiosInt15Handler.cs
```diff
@@ -20,14 +23,25 @@ public class SystemBiosInt15Handler : InterruptHandler {
...
+        uint bytes = (uint)(_state.CX * 2);
+        uint data = MemoryUtils.ToPhysicalAddress(_state.ES, _state.SI);
+        long source = _memory.UInt32[data + 0x12] & 0x00FFFFFF + _memory.UInt8[data + 0x16] << 24;
```
kevinferrare, line 37: Looks like a struct in memory. Maybe having a full class would make the intent a bit clearer?

src/Spice86.Core/Emulator/InterruptHandlers/Bios/SystemBiosInt15Handler.cs
maximilien-noal: Properties are not supposed to have any side effect (calls, calculations, touching other properties, ...).
This is supposed to modify the wrapping, so this is why it is a method.
It's not implemented, so I'll remove it.

src/Spice86.Core/Emulator/Memory/ReservedBlock.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace Spice86.Core.Emulator.Memory;
+
+/// <summary>
+/// Describes a reserved block of conventional memory.
+/// </summary>
+public sealed class ReservedBlock : IEquatable<ReservedBlock> {
```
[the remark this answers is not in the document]
maximilien-noal: Unused class. Good catch ! I removed it.

src/Spice86.Core/Emulator/Memory/Memory.cs
maximilien-noal: Code removed.

src/Spice86.Core/Emulator/InterruptHandlers/Bios/SystemBiosInt15Handler.cs
maximilien-noal: Code removed.

src/Spice86.Core/Emulator/InterruptHandlers/Bios/SystemBiosInt15Handler.cs
maximilien-noal: Code removed.

PR #312, opened 2023-05-18
A20 gate (was: DosMemoryManager: don't allocate MCB beyond 1MB) | WIP

maximilien-noal:
This makes the Syndicate INTRO.EXE run until it tries to run DOS4GW.EXE, with the following message:

>Stub exec failed:
C:\dos4gw.exe
No such file or directory

Previously, the emulator would eventually throw an ArgumentOutRangeException when the PspSegment property tried to perform 'GetUint16(PspSegmentFieldOffset)' with a value that was out of range for conventional memory.

I think this is a good enough fix for these reasons:

1. DOS is expected to live in the first 1 MB, so the validation of MCBs is the responsibility of the DosMemoryControlBlock.Next method.
2. We use the Try... pattern to avoid an allocation if possible and to avoid the exception
3. We protect against the exception by checking the base address. Surely, if the base address is out of range or at the very end, then the access to the PspSegment property or any other will fail. Any better check ?
4. When the request is determined to be out of range, we return false, and set the output variable to 'this'. This avoid any more code change. We don't set the output variable to null, not return null, because this would involve more changes than strictly necessary.

But I'm certainly not against comments or improvements.

Especially since the allocation is behind a Next() method... Would it make more sense to make it more obvious ? With a better named method, perhaps ?

(For the record, DOS4GW.EXE is in the directory, and the DosFileManager returns the correct casing)

My aim is not to make Syndicate work (even if it would be cool), it's just an issue that I encountered and wasn't really aware of until I looked into the code. Seemed to me that it makes sense to change this code, in line with the previous introduction of the Memory Bus.

EDIT:
The real fix is to implement an A20 gate.

Back to the drawing board !


maximilien-noal:
The real fix is to implement an A20 gate.

Back to the drawing board !

maximilien-noal:
Dropped, preparing a new PR for the A20 gate.

PR #315, opened 2023-05-19
Feature: A20 Gate emulation

### LINKED ISSUE #313: Feature request: A20 gate emulation
See the XMS branch.

maximilien-noal:
Closes #313  and fixes the MCB class and DosMemoryManager so no MCB can be allocated beyond ~1 MB.

This makes Syndicate run until it tries to run DOS4GW (the file is found, but fixing Syndicate is not the purpose of this PR).


src/Spice86.Core/Emulator/Memory/Memory.cs
```diff
@@ -333,11 +359,13 @@ public class Memory {
     }
 
     private void Write(uint address, byte value) {
+        address &= A20AddressMask;
```
kevinferrare, line 362: Would be nice if performance is not too much impacted by this to also do a modulo memory size (which would differ between a20 enabled or not?) since this is how it works in hardware and some games may use this obscure feature.
If you choose to implement this, it could make sense to move all this to a class that just computes the physical address based on this and A20. Could be useful for rewritten code as well to have such class.

src/Spice86.Core/Emulator/OperatingSystem/Structures/DosMemoryControlBlock.cs
```diff
@@ -79,11 +79,18 @@ public class DosMemoryControlBlock : MemoryBasedDataStructureWithBaseAddress {
     public bool IsValid => IsLast || IsNonLast;
 
     /// <summary>
-    /// Returns the next MCB in the MCB chain.
+    /// Returns the next MCB in the next MCB in chain, along with a boolean status.
     /// </summary>
-    /// <returns></returns>
-    public DosMemoryControlBlock Next() {
-        return new DosMemoryControlBlock(Memory, BaseAddress + MemoryUtils.ToPhysicalAddress((ushort)(Size + 1), 0));
+    /// <param name="next">The output variable containing the next MCB, or this MCB.</param>
```
kevinferrare, line 84: Why not keeping the same semantic (next) but returning null when the end of memory is reached?

src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
```diff
@@ -46,13 +50,43 @@ public class Keyboard : DefaultIOPortHandler {
     public override byte ReadByte(int port) {
         byte? scancode = LastKeyboardInput.ScanCode;
         if (scancode == null) {
-            return 0;
+            scancode = 0;
+        }
+        
+        return port switch {
+            KeyboardPorts.IOPort => scancode.Value,
```
kevinferrare, line 57: Shouldnt we read the keyboard buffer just when we are requested to read scan code?

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardPorts.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// <summary>
+    /// Port that can be used by the CPU to read the status register of the keyboard controller.
+    /// </summary>
+    public const byte StatusRegister = 0x64;
+
+    /// <summary>
+    /// If the CPU writes to port 0x64, the byte is interpreted as a command byte. 
+    /// </summary>
+    public const byte CommandPort = 0x64;
```
kevinferrare, line 21: It's a bit confusing to have 2 names for the same port value. Do we have other devices like this?

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardPorts.cs
```diff
@@ -0,0 +1,22 @@
...
+
+/// <summary>
+/// IDs of keyboard ports for both read and write. <br/>
+/// https://www.win.tue.nl/~aeb/linux/kbd/scancodes-11.html
+/// </summary>
+public static class KeyboardPorts {
+    /// <summary>
+    /// Port used by the CPU to read the input buffer, or write to the output buffer, of the keyboard controller.
+    /// </summary>
+    public const byte IOPort = 0x60;
```
kevinferrare, line 11: How about calling this Data like they do here? https://bochs.sourceforge.io/techspec/PORTS.LST

src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
```diff
@@ -46,13 +50,43 @@ public class Keyboard : DefaultIOPortHandler {
...
+        switch (port) {
+            // the byte is interpreted as a data byte
+            case KeyboardPorts.IOPort:
+                if (Command == KeyboardCommand.SetOutputPort) {
+                    _machine.Memory.IsA20GateEnabled = (value & 2) > 0;
+                    Command = KeyboardCommand.None;
+                }
+                break;
+            case KeyboardPorts.CommandPort:
+                Command = value switch {
```
kevinferrare, line 76: Wouldnt it be clearer like that?
```
if(value == KeyboardValues.SetKeyboardCommand) {
  Command = KeyboardCommand.SetCommand;
}
```

src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
```diff
@@ -46,13 +50,43 @@ public class Keyboard : DefaultIOPortHandler {
...
         byte? scancode = LastKeyboardInput.ScanCode;
         if (scancode == null) {
-            return 0;
+            scancode = 0;
+        }
+        
+        return port switch {
+            KeyboardPorts.IOPort => scancode.Value,
+            // keyboard not locked, self-test completed.
+            KeyboardPorts.StatusRegister => 0x14,
```
kevinferrare, line 59: Since each bit has a meaning, shouldnt we write this like that? Maybe in the future we could have variable bits with a more refined implementation.
```
const byte SYSTEM_TEST_STATUS_MASK = 1<<2;
const byte KEYBOARD_ENABLE_STATUS_MASK = 1<<4;
KeyboardPorts.StatusRegister => SYSTEM_TEST_STATUS_MASK | KEYBOARD_ENABLE_STATUS_MASK;

```

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardPorts.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/Memory/Memory.cs
maximilien-noal: Done (putting the A20 Gate into a class) and done (using modulo on top). :)

src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
maximilien-noal: Not really. Maybe VGA ports ?
This is the same as here:
https://github.com/sarah-walker-pcem/pcem/blob/dev/src/keyboard/keyboard_at.c#L314

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardPorts.cs
maximilien-noal: Not really. Maybe VGA ports ?
This is the same as here:
https://github.com/sarah-walker-pcem/pcem/blob/dev/src/keyboard/keyboard_at.c#L314

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardPorts.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
maximilien-noal: Done, and ports corrected by looking at the PCem code for the keyboard.

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardPorts.cs
JorisVanEijden: Yeah, I have VGA ports with different names for reading and writing when they're being used in read-only or write-only context.
To me different names for the same port in different contexts is fine.

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardCommand.cs
```diff
@@ -0,0 +1,17 @@
...
+public enum KeyboardCommand : byte {
+    /// <summary>
+    /// No command set.
+    /// </summary>
+    None = 0x0,
+    
+    /// <summary>
+    /// Write to the output port
+    /// </summary>
+    SetOutputPort = 0xD1,
```
kevinferrare, line 16: I think there are commands that are missing for this to work, see how it's called from the programmer side: https://wiki.osdev.org/A20_Line#Enabling
We would miss 0xAD, 0xD0, 0xAE

src/Spice86.Core/Emulator/Memory/Memory.cs
JorisVanEijden: Actually, the modulo is an effect of the AND performed by the hardware, which comes down to (address &= ((2**addressBits) -1))

Example: 
Using 20 bits (A20 disabled):
2**20  = 0x100000
0x100000 - 1 = 0xFFFFF
0x100123 & 0xFFFFF = 0x123

Using 21 bits (A20 enabled):
2**21 = 0x200000
0x200000 - 1 = 0x1FFFFF
0x100123 & 0x1FFFFF = 0x100123

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardCommand.cs
maximilien-noal: Indeed, and there is also a simialr command in the BIOS to be implemented.

src/Spice86.Core/Emulator/Memory/Memory.cs
maximilien-noal: Done (calculations simplified).

src/Spice86.Core/Emulator/Devices/Input/Keyboard/KeyboardCommand.cs
maximilien-noal: BIOS A20 Gate support implemented. Keyboard missing A20 Gate related commands implemented.

PR #320, opened 2023-05-28
A new VGA videocard implementation.

JorisVanEijden:
It features a BIOS that is ported from an actual VGA BIOS (seabios).

Has a custom single renderer that can deal with all modes.

Adds support for all "VGA Modes". Text, Graphics, CGA, 256-color, etc.


src/Spice86.Core/Emulator/Devices/Video/VgaIoPortHandler.cs
```diff
@@ -0,0 +1,354 @@
...
+            Ports.MiscOutputWrite,
+            Ports.SequencerAddress,
+            Ports.SequencerData,
+            Ports.CgaModeControl,
+            Ports.CgaColorSelect
+        };
+
+    /// <inheritdoc />
+    public override void InitPortHandlers(IOPortDispatcher ioPortDispatcher) {
+        foreach (int port in HandledPorts) {
```
kevinferrare, line 79: Would be nice to have IOPortDispatcher supporting adding a list, could be used by other devices (most?)

src/Spice86.Core/Emulator/Devices/Video/VideoMemory.cs
```diff
@@ -1,32 +1,260 @@
...
+        Planes = new byte[4, 0x20000];
+        _latches = new byte[4];
+        Size = 0x20000;
     }
+
+    /// <inheritdoc />
+    public uint Size { get; }
+
+    /// <inheritdoc />
     public byte Read(uint address) {
```
kevinferrare, line 30: If I understand well this is done for each pixel displayed.
I am wondering if it would make sense to have a cache for this:
 - first read would populate the cache
 - second read would read from the cache
 writing at the planes / registers / latches would invalidate the cache entries.

How is performance compared to existing solution?

src/Spice86.Core/Emulator/Devices/Video/Registers/Register8.cs
```diff
@@ -0,0 +1,54 @@
+namespace Spice86.Core.Emulator.Devices.Video.Registers;
+
+using System.Runtime.CompilerServices;
+
+public class Register8 {
```
kevinferrare, line 5: Wondering if this could be mutualized with other devices we have. Probably not easily with the cpu registers as they are 32bits and provide access to 16 and 8 bits.

src/Spice86/ViewModels/MainWindowViewModel.cs
```diff
@@ -184,7 +184,7 @@ public sealed partial class MainWindowViewModel : ObservableObject, IGui, IDispo
         IVideoBufferViewModel videoBuffer = new VideoBufferViewModel(videoCard, scale, bufferWidth, bufferHeight, address, VideoBuffers.Count, isPrimaryDisplay);
         Dispatcher.UIThread.Post(
             () => {
-                if(!VideoBuffers.Any(x => x.Address == videoBuffer.Address)) {
+                if(VideoBuffers.All(x => x.Address != videoBuffer.Address)) {
```
maximilien-noal, line 187: Any should be faster than All, since it stops at the first item where the condition is not true ?

src/Spice86/ViewModels/MainWindowViewModel.cs
```diff
@@ -395,6 +395,18 @@ await MessageBox.Avalonia.MessageBoxManager.GetMessageBoxStandardWindow("", "Ple
...
     }
+    
+    public void SetResolution(int width, int height) {
+        Dispatcher.UIThread.Post(() => {
+            _isSettingResolution = true;
+            DisposeBuffers();
+            VideoBuffers = new();
+            Width = width;
+            Height = height;
+            AddBuffer(_videoCard, 0xA0000, 1, width, height, true);
```
maximilien-noal, line 406: Is the usage of 0xA0000 here work with video modes that don't start at that address ?

src/Spice86.Core/Emulator/Gdb/GdbCustomCommandsHandler.cs
```diff
@@ -380,7 +380,8 @@ public class GdbCustomCommandsHandler {
                     return _gdbIo.GenerateMessageToDisplayResponse($"Buffer already exists: {existing}");
                 }
 
-                gui?.AddBuffer(_machine.VgaCard, address, scale, resolution[0], resolution[1]);
+                // Todo: Fix this?
+                // gui?.AddBuffer(_machine.VgaIoPortHandler, address, scale, resolution[0], resolution[1]);
```
kevinferrare, line 384: I think you can remove the feature altogether, it was never very useful

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -411,7 +411,27 @@ public class DosInt21Handler : InterruptHandler {
...
+        // Set up values for bios call
+        _state.BL = 0x0F; // white on black
+        _state.ES = segment;
+        _state.BP = offset;
+        _state.CX = (ushort)str.Length;
+        _state.AL = 0x00; // no attribute, no cursor movement
+        _state.BH = _machine.Bios.CurrentVideoPage;
+        _state.DX = _machine.Bios.GetCursorPosition(_state.BH);
+        _state.AH = 0x13; // Write string
+        _machine.CallbackHandler.Run(0x10);
```
kevinferrare, line 430: Would be nice to have an accessible public method in video bios for this instead.

src/Spice86.Core/Emulator/InterruptHandlers/VGA/VgaBios.cs
```diff
@@ -0,0 +1,1496 @@
...
+            SegmentedAddress font = GetFontAddress(character);
+            if (MemCmp(lines, font.Segment, font.Offset, characterHeight) == 0) {
+                return new CharacterPlusAttribute(character, foregroundAttribute, false);
+            }
+        }
+
+        return new CharacterPlusAttribute((char)0, 0, false);
+    }
+
+    private int MemCmp(IReadOnlyList<byte> bytes, ushort segment, ushort offset, int length) {
```
kevinferrare, line 1186: Could be moved to a common place maybe, not sure if it makes sense to have this reuseable?

src/Spice86.Core/Emulator/Gdb/GdbCustomCommandsHandler.cs
maximilien-noal: It would simplify the UI code.

src/Spice86.Core/Emulator/InterruptHandlers/VGA/Records/VideoMode.cs
```diff
@@ -0,0 +1,6 @@
+namespace Spice86.Core.Emulator.InterruptHandlers.VGA.Records;
+
+/// <summary>
+///
```
maximilien-noal, line 4: Missing documentation

src/Spice86.Core/Emulator/InterruptHandlers/VGA/Records/CursorPosition.cs
```diff
@@ -0,0 +1,6 @@
+namespace Spice86.Core.Emulator.InterruptHandlers.VGA.Records;
+
+/// <summary>
+///     Represents a cursor position.
+/// </summary>
+public record struct CursorPosition(int X, int Y, int Page);
```
maximilien-noal, line 6: Missing documentation for Page. What is a page in this context ?

src/Spice86.Core/Emulator/Devices/Video/Registers/Enums/CrtControllerRegister.cs
```diff
@@ -0,0 +1,29 @@
+namespace Spice86.Core.Emulator.Devices.Video.Registers.Enums;
+
+public enum CrtControllerRegister {
```
maximilien-noal, line 3: Missing documentation that explains what each value means.

src/Spice86.Core/Emulator/Devices/Video/Registers/Enums/AttributeControllerRegister.cs
```diff
@@ -0,0 +1,11 @@
+namespace Spice86.Core.Emulator.Devices.Video.Registers.Enums;
+
+public enum AttributeControllerRegister {
```
maximilien-noal, line 3: Missing documentation that explains what each value means.

src/Spice86.Core/Emulator/Devices/Video/VideoMemory.cs
JorisVanEijden: This is for reading video memory from the application side, not the card/renderer side.

This is mostly used for reading exisiting graphics, for instance when moving a mouse pointer over them. So not used a often as writing to video memory.

Performance is quite good actually. Not worth building a complicated caching mechanism for in my opinion.

src/Spice86.Core/Emulator/Devices/Video/VgaIoPortHandler.cs
JorisVanEijden: Could be, I'm not sure it belongs in this PR or even if it makes any difference in performance or ease-of-development.

src/Spice86.Core/Emulator/Devices/Video/Registers/Register8.cs
JorisVanEijden: I'm not sure what the benefits would be.

src/Spice86/ViewModels/MainWindowViewModel.cs
JorisVanEijden: `Any` also stops at the 1st entry that matches the condition doesn't it?

So I think it't just a syntactic difference between "if not any of the buffers have this address" and "if all of the buffers don't have this address".

src/Spice86/ViewModels/MainWindowViewModel.cs
JorisVanEijden: Actually, it completely ignores that address. Also in the existing code.

I'd have to think a bit on how to make it so it can take an arbitrary address rather than the one programmed into the registers.

src/Spice86/ViewModels/MainWindowViewModel.cs
maximilien-noal: My bad. The performance is the same.

src/Spice86.Core/Emulator/Gdb/GdbCustomCommandsHandler.cs
JorisVanEijden: Phew, that's good to hear. Although it should not be too hard to re-implement something like this after everything has settled.

src/Spice86/ViewModels/MainWindowViewModel.cs
maximilien-noal: Removing the feature of adding/remove arbitrary video buffers from the UI and IGui is fine too.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
JorisVanEijden: I'm not sure. I made it this way so anything else can override / take over INT 10 and it'd still work. A hard dependency between DOS and the current VGA bios does not sound that beneficial.

src/Spice86.Core/Emulator/InterruptHandlers/VGA/VgaBios.cs
JorisVanEijden: Since this is a port of an existing bios I tried to make it as stand-alone and independent as possible.
Nothing should have to depend on it and it shouldn't have to depend on anything.

Which is also why it reads and writes the I/O ports rather than manipulating an instance of the VideoState directly.

src/Spice86.Core/Emulator/Devices/Video/Registers/Enums/AttributeControllerRegister.cs
JorisVanEijden: It's just the names of the registers. But I added docs to shut the analyzer up ;)

src/Spice86.Core/Emulator/Devices/Video/Registers/Enums/CrtControllerRegister.cs
JorisVanEijden: It's just the names of the registers. But I added docs to shut the analyzer up ;)

src/Spice86.Core/Emulator/InterruptHandlers/VGA/Records/CharacterPlusAttribute.cs
```diff
@@ -0,0 +1,6 @@
+namespace Spice86.Core.Emulator.InterruptHandlers.VGA.Records;
+
+/// <summary>
+///     Represents a character with an attribute.
+/// </summary>
+public record struct CharacterPlusAttribute(char Character, byte Attribute, bool UseAttribute);
```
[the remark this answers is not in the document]
JorisVanEijden: Docs added

src/Spice86.Core/Emulator/InterruptHandlers/VGA/Records/CursorPosition.cs
JorisVanEijden: Docs added

src/Spice86.Core/Emulator/InterruptHandlers/VGA/Records/VideoMode.cs
JorisVanEijden: Docs added

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
kevinferrare: I suppose you are thinking about games that may provide their own implementation of INT 10 or loading a real VGA bios beside emulated DOS.
The thing is we are not able to call ASM code from C# (and it is a difficult problem because you don't have a guarantee the ASM will be nice and return to you).

JorisVanEijden: I have refactored the VgaBios to have a public interface usable by outside code.
And an interrupt interface that only handles interrupt functions.

maximilien-noal:
There is a regression with Dune: menu animations don't show anymore.

There are timing sensitive animations in this game, and they are most notable in the menu used during gameplay.

They show up on the master branch.

I can provide further information about this issue, if required.

PR #321, opened 2023-05-31
fix: improve performance by not creating a full logger only to check its enabled status

Arcnor:
Did a few tests using my current example (Detroit's intro). With this change, and at least in debug, I got from 30 seconds to render up to the first "pixel fade" to only 20, with logging disabled.

With logging enabled more improvements can probably be made (because the hotspots will now be the different `Write`s). The main culprit is `AddProperties` on `GetLogger`, and inside that one the main issue is `String` concatenation while creating the segmented address. I'm sure this can be cached instead of recreated for every single logging event.


maximilien-noal:
Nice catch !

Looks good to me.

kevinferrare (approved):

PR #335, opened 2023-06-17
fix: video renderer Argument or IndexOutOfRangeException crashes

### LINKED ISSUE #332: VGA Renderer: IndexOutOfRangeException with some programs in DrawTextMode (tested with Hero Quest's intro)
**Describe the bug**

Some programs, such as Cadaver's INTRO.EXE and Hero Quest's INTRO.EXE make Renderer.cs throw an IndexOutOfRangeException:

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Spice86.Core.Emulator.Devices.Video.Renderer.DrawTextMode(Span`1 frameBuffer, Int32& destinationAddress, Byte plane0, Byte plane1, Int32 scanline) in /home/max/repos/Spice86/src/Spice86.Core/Emulator/Devices/Video/Renderer.cs:line 249
   at Spice86.Core.Emulator.Devices.Video.Renderer.Render(Span`1 frameBuffer) in /home/max/repos/Spice86/src/Spice86.Core/Emulator/Devices/Video/Renderer.cs:line 136
   at Spice86.Core.Emulator.Devices.Video.VgaCard.Render(Span`1 buffer) in /home/max/repos/Spice86/src/Spice86.Core/Emulator/Devices/Video/VgaCard.cs:line 48
   at Spice86.ViewModels.VideoBufferViewModel.<Draw>b__31_0() in /home/max/repos/Spice86/src/Spice86/ViewModels/VideoBufferViewModel.cs:line 161
   at Spice86.ViewModels.VideoBufferViewModel.DrawThreadMethod() in /home/max/repos/Spice86/src/Spice86/ViewModels/VideoBufferViewModel.cs:line 54

```

This happens at the beginning of the intro.

The code line number 249 ("frameBuffer[destinationAddress++] = pixel;") triggers the error:

```
[...]
        // The 8 pixels to render this line come from the font which is stored in plane 2.
        byte fontByte = _memory.Planes[2, fontAddress + scanline];
        for (int x = 0; x < _state.SequencerRegisters.ClockingModeRegister.DotsPerClock; x++) {
            uint pixel = (fontByte & 0x80 >> x) != 0 ? foreGroundColor : backGroundColor;
            frameBuffer[destinationAddress++] = pixel;
        }
[...]
```

```
backGroundColor = {uint} 4278190250 0xFF0000AA
destinationAddress = {int} 256001 0x3E801
fontAddress = {int} 576 0x240
fontByte = {byte} 24 0x18
foreGroundColor = {uint} 4278233600 0xFF00AA00
frameBuffer = {Span<uint>} System.Span<UInt32>[256000]
index = {int} 2 0x2
pixel = {uint} 4278190250 0xFF0000AA
plane0 = {byte} 18 0x12
plane1 = {byte} 18 0x12
scanline = {int} 7 0x7
x = {int} 7 0x7
```

**To Reproduce**

1. Run Spice with the following command line arguments:

```
-d false --Ems -s
```

2. Program -> Start Program... -> Select INTRO.EXE
3. Wait for the error.

**Expected behavior**

The Hero Quest INTRO should run.

**Screenshots**

image

 - OS: Fedora Workstation
 - Version: master branch commit 9c9a137

**Additional context**

### LINKED ISSUE #333: VGA Renderer: IndexOutRangeException in some programs in GetDacPaletteColor (tested with Duke Nukem II)
**Describe the bug**

The start of Duke Nukem II produces a different kind of IndexOutRangeException in Renderer.cs on line 307 ("int fromPaletteRam6Bits = _state.AttributeControllerRegisters.InternalPalette[index];") :

```
private uint GetDacPaletteColor(int index) {
        switch (_state.AttributeControllerRegisters.AttributeControllerModeRegister.PixelWidth8) {
            case true:
                return _state.DacRegisters.ArgbPalette[index];
            default: {
                int fromPaletteRam6Bits = _state.AttributeControllerRegisters.InternalPalette[index];
                int bits0To3 = fromPaletteRam6Bits & 0b00001111;
                int bits4And5 = _state.AttributeControllerRegisters.AttributeControllerModeRegister.VideoOutput45Select
                    ? _state.AttributeControllerRegisters.ColorSelectRegister.Bits45 << 4
                    : fromPaletteRam6Bits & 0b00110000;
                int bits6And7 = _state.AttributeControllerRegisters.ColorSelectRegister.Bits67 << 6;
                int dacIndex8Bits = bits6And7 | bits4And5 | bits0To3;
                int paletteIndex = dacIndex8Bits & _state.DacRegisters.PixelMask;
                return _state.DacRegisters.ArgbPalette[paletteIndex];
            }
        }
    }
```

**To Reproduce**
1. Run Spice86 with this command line:
```
-d false --Ems -s
```
2. Program -> Start Program... -> Select NUKEM2.EXE

**Expected behavior**

Duke Nukem II should not produce this error.

**Screenshots**

image

**Desktop (please complete the following information):**
 - OS: Fedora
 - Version: master commit b3a3de7

**Additional context**

state of local variables when the error occured:

```
bits0To3 = {int} 0 0x0
bits4And5 = {int} 0 0x0
bits6And7 = {int} 0 0x0
dacIndex8Bits = {int} 0 0x0
fromPaletteRam6Bits = {int} 0 0x0
index = {int} 197 0xC5
paletteIndex = {int} 0 0x0
```

The point at which the exception occurs is not always the same.

maximilien-noal:
Fixes #333 
Fixes #332

The try...catch(ArgumentOutOfRangeException) { } variant killed performance instantly with Duke Nukem II. 
Sound stopped, video stopped, everthing stopped forever.

Tested successfully on:

- Hero Quest
- Duke Nukem II
- Cadaver intro

### DIFF
--- a/src/Spice86.Core/Emulator/Devices/Video/Renderer.cs
+++ b/src/Spice86.Core/Emulator/Devices/Video/Renderer.cs
@@ -246,7 +246,13 @@ public class Renderer : IVgaRenderer {
         byte fontByte = _memory.Planes[2, fontAddress + scanline];
         for (int x = 0; x < _state.SequencerRegisters.ClockingModeRegister.DotsPerClock; x++) {
             uint pixel = (fontByte & 0x80 >> x) != 0 ? foreGroundColor : backGroundColor;
-            frameBuffer[destinationAddress++] = pixel;
+            TryWriteElementAt(frameBuffer, destinationAddress++, pixel);
+        }
+    }
+
+    private static void TryWriteElementAt(Span<uint> span, int index, uint value) {
+        if (index < span.Length) {
+            span[index] = value;
         }
     }
 
@@ -264,12 +270,12 @@ public class Renderer : IVgaRenderer {
         // First 4 pixels are created from planes 0 and 2.
         for (int bitNr = 6; bitNr >= 0; bitNr -= 2) {
             int index = (plane2 >> bitNr & 3) << 2 | plane0 >> bitNr & 3;
-            frameBuffer[destinationAddress++] = GetDacPaletteColor(index);
+            TryWriteElementAt(frameBuffer, destinationAddress++, GetDacPaletteColor(index));
         }
         // Then 4 pixels are created from planes 1 and 3.
         for (int bitNr = 6; bitNr >= 0; bitNr -= 2) {
             int index = (plane3 >> bitNr & 3) << 2 | plane1 >> bitNr & 3;
-            frameBuffer[destinationAddress++] = GetDacPaletteColor(index);
+            TryWriteElementAt(frameBuffer, destinationAddress++, GetDacPaletteColor(index));
         }
     }
 
@@ -278,7 +284,7 @@ public class Renderer : IVgaRenderer {
         // outputting 8 pixels.
         for (int bitNr = 7; bitNr >= 0; bitNr--) {
             int index = (plane3 >> bitNr & 1) << 3 | (plane2 >> bitNr & 1) << 2 | (plane1 >> bitNr & 1) << 1 | plane0 >> bitNr & 1;
-            frameBuffer[destinationAddress++] = GetDacPaletteColor(index);
+            TryWriteElementAt(frameBuffer, destinationAddress++, GetDacPaletteColor(index));
         }
     }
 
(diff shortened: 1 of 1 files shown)


JorisVanEijden:
I think there may be an underlying problem/bug here, if adding a `catch` has any effect on performance.
In my measurements of a release build with `catch` statement, I get an average of 2.0-2.1 milliseconds per frame. Without the `catch` statement it's also between 2.0 and 2.1 milliseconds.
In a debug build I do see a very slight difference where `catch` is 0.05-0.1 milliseconds per frame slower. Of course this only means the delay loop has to wait a little bit less.

If the exception is thrown more than twice during the start of the game, this also indicates a different problem, because as far as I see it only switches resolution twice.

src/Spice86.Core/Emulator/Devices/Video/Renderer.cs
```diff
@@ -289,30 +295,42 @@ public class Renderer : IVgaRenderer {
...
-                int fromPaletteRam6Bits = _state.AttributeControllerRegisters.InternalPalette[index];
-                int bits0To3 = fromPaletteRam6Bits & 0b00001111;
-                int bits4And5 = _state.AttributeControllerRegisters.AttributeControllerModeRegister.VideoOutput45Select
-                    ? _state.AttributeControllerRegisters.ColorSelectRegister.Bits45 << 4
-                    : fromPaletteRam6Bits & 0b00110000;
-                int bits6And7 = _state.AttributeControllerRegisters.ColorSelectRegister.Bits67 << 6;
-                int dacIndex8Bits = bits6And7 | bits4And5 | bits0To3;
-                int paletteIndex = dacIndex8Bits & _state.DacRegisters.PixelMask;
-                return _state.DacRegisters.ArgbPalette[paletteIndex];
+                int? fromPaletteRam6BitsOrNull = GetElementAtOrNull(_state.AttributeControllerRegisters.InternalPalette, index);
```
JorisVanEijden, line 320: This appears to be a fix for a whole different problem. One that has nothing to do with resoluation changes.
If the `AttributeControllerModeRegister.PixelWidth8` bit is False, then "index" should be a 4-bit number.

I have so far been unable to reproduce this situation so I don't know where the higher index number comes from. But I think finding and fixing that would be better than to hide the bug this way.

Perhaps this is also a timing issue, where the register has already been disabled while it is still reading 8-bit data.
In that case I should see if we maybe need to latch the register at the start of the frame or the start of the row.

maximilien-noal: I'm way out of my depth, I'm afraid. What do you mean by latching ?

PR #336, opened 2023-06-18
Fix video crashes

JorisVanEijden:
- Prevent overflow of internal palette.
- Ignore framebuffer overflow (can happen when resolution changes mid-frame)


maximilien-noal:
Everything is fine with it. Hero Quest, Krondor, Dune, Dune 2, Duke Nukem I, Duke Nukem II, ... Great job !

PR #346, opened 2023-06-23
Refactoring/service container

### LINKED ISSUE #338: Feature request: make the emulator code more testable. Refactor the Machine especially.

maximilien-noal:
### Description of Changes

Cleaned up startup code. For example, we only parse the command line once.
Uses [StrongInject](https://github.com/YairHalberstadt/stronginject) for dependency injection of services.

Another option would have been [StashBox](https://github.com/z4kn4fein/stashbox/blob/master/docs/docs/getting-started/introduction.md). But StrongInject doesn't force the usage of interfaces. Interfaces are great for IoC, decoupling, and unit testing. Not so much as busy work to make a DI framework happy.

The Startup code is now a service.
The idea is that in the future, the Top Level Container will contain all services for the UI.

Another container will be used for the emulator layer. 

The Microsoft nuget package for DI has been removed, as it depends on the GenericHost for DI, and that cannot run at the same time as the AvaloniaUI host. Calling one or the other is a blocking call.
On the other hand, StrongInject is injected at compile time so it has no such conflict.
It's also blazing fast.

Machine has been divided into subsystems, GDB specific methods moved to the GDB namespace.

### Rationale behind Changes

Make the code more testable.
Closes #338 
But of course, making the code more testable is a work in progress.
The primary focus on this PR was to make the Machine into a more SOLID class, not a God class.

### Suggested Testing Steps

Already supported games were tested.
No regression found.

Logging still works, even if we now use a TopLevelContainer, including the dynamic log level.


JorisVanEijden:
I don't really understand the rationale for moving the intterrupt and port handlers deeper into the machine with these "subsystems".
The Machine still contains all of them so it is still just as much a God class that is passed everywhere, just with a lot more Law of Demeter violations.

maximilien-noal:
This is only but a step.

The next step would be to decouple with interfaces between the machine and its subsystems.

src/Spice86/Program.cs
```diff
@@ -34,37 +30,9 @@ public static void RunWithOverrides<T>(string[] args, string expectedChecksum) w
...
-    /// <returns>The built <see cref="AppBuilder"/> instance.</returns>
-    public static AppBuilder BuildAvaloniaApp()
-    {
-        return AppBuilder.Configure<App>()
-            .UsePlatformDetect()
-            .LogToTrace();
+    public static void Main(string[] args) {
+        using TopLevelContainer topLevelContainer = new();
+        Owned<Startup> startup = topLevelContainer.Resolve();
+        startup.Value.StartApp(args);
```
JorisVanEijden, line 36: I'm still reading the StrongInject docs, but should this not be
```suggestion
        topLevelContainer.Run(startup => startup.Start(args));
```
?

maximilien-noal:
I'll see if I can make the IoC a reality without making the PR too big.

src/Spice86.Shared/Interfaces/IMainWindowViewModel.cs
```diff
@@ -6,9 +6,8 @@
 /// GUI of the emulator.<br/>
 /// Displays the content of the video ram (when the emulator requests it) <br/>
 /// Communicates keyboard and mouse events to the emulator <br/>
-/// This is the MainWindowViewModel.
 /// </summary>
-public interface IGui {
+public interface IMainWindowViewModel {
```
JorisVanEijden, line 10: I feel an interface should be named for what it _does_, not for what it _is_. So this rename makes no sense to me.

src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
```diff
@@ -114,7 +111,7 @@ public class Timer : DefaultIOPortHandler {
         }
 
         if (_vgaScreenRefreshCounter.ProcessActivation(cycles)) {
-            _vgaCard?.UpdateScreen();
+            _machine.VideoSubsystem.VgaCard.UpdateScreen();
```
JorisVanEijden, line 114: This is the reverse of the direction I'd like to see.
Ideally these classes do **not** get a Machine injected, but only the things they depend on, through interfaces.
We want to put less in the Machine, not more.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
```diff
@@ -11,8 +11,8 @@
 public class BiosKeyboardInt9Handler : InterruptHandler {
     private readonly Keyboard _keyboard;
 
-    public BiosKeyboardInt9Handler(Machine machine, ILoggerService loggerService) : base(machine, loggerService) {
-        _keyboard = machine.Keyboard;
+    public BiosKeyboardInt9Handler(Machine machine, Keyboard keyboard, ILoggerService loggerService) : base(machine, loggerService) {
```
JorisVanEijden, line 14: Yes, this is better, depend on the keyboard itself and not depend on the machine to provide a keyboard.
Now it just needs to depend on an IKeyboardDevice interface rather than a concrete implementation of the KeyBoard class.

src/Spice86/Startup.cs
```diff
@@ -1,54 +1,80 @@
...
-    /// Sets the logging level based on the command line arguments.
-    /// </summary>
-    /// <param name="serviceProvider">The <see cref="ServiceProvider"/> instance used to retrieve the <see cref="ILoggerService"/>.</param>
-    /// <param name="commandLineArgs">The command line arguments.</param>
-    private static void SetLoggingLevel(ServiceProvider serviceProvider, string[] commandLineArgs) {
-        ILoggerService? loggerService = serviceProvider.GetService<ILoggerService>();
-        if (loggerService is null) {
-            return;
+    
+    public void StartApp(string[] args) {
```
JorisVanEijden, line 23: I think we'd better inject the configuration here, rather than the args.

src/Spice86/Program.cs
```diff
@@ -34,37 +30,9 @@ public static void RunWithOverrides<T>(string[] args, string expectedChecksum) w
...
-    /// Configures and builds an Avalonia application instance.
-    /// </summary>
-    /// <returns>The built <see cref="AppBuilder"/> instance.</returns>
-    public static AppBuilder BuildAvaloniaApp()
-    {
-        return AppBuilder.Configure<App>()
-            .UsePlatformDetect()
-            .LogToTrace();
+    public static void Main(string[] args) {
+        using TopLevelContainer topLevelContainer = new();
```
JorisVanEijden, line 34: If we resolve the configuration first, we can inject it into the container and have it available everywhere.
See https://github.com/YairHalberstadt/stronginject#providing-registrations-at-runtime-or-integrating-with-other-ioc-containers

And if we make the Configuration an `IConfiguration`  we can use commandline, jsonfile, environment variable, etc. `IConfigurationBuilder`s

JorisVanEijden:
> This is only but a step.
> 
> The next step would be to decouple with interfaces between the machine and its subsystems.

I'm more confused. The Machine class doesn't use most of these "subsystems". So why should there be interfaces between them?

JorisVanEijden:
> I'll see if I can make the IoC a reality without making the PR too big.

I'm not sure what this is a response to. 
I think the choice of and start with StrongInject could be a good one. We can do that step by step.

src/Spice86/Program.cs
maximilien-noal: I was about to use IConfiguration, but feared if was too much of a change at once.

I don't know how well it will mix the [Option] tags from Configuration.cs, which is exposed for command line parsing and to provide commandeline help.

maximilien-noal:
> > This is only but a step.
> > The next step would be to decouple with interfaces between the machine and its subsystems.
> 
> I'm more confused. The Machine class doesn't use most of these "subsystems". So why should there be interfaces between them?

Sorry,  you're right. The Machine itself right now is still confusing my brain.

src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
maximilien-noal: To depend on the VgaCard directly would produce a NullReferenceException later. At the time of initialization, the VgaCard does not exist yet.

src/Spice86.Shared/Interfaces/IMainWindowViewModel.cs
maximilien-noal: Reversed.

maximilien-noal:
> I don't really understand the rationale for moving the intterrupt and port handlers deeper into the machine with these "subsystems". The Machine still contains all of them so it is still just as much a God class that is passed everywhere, just with a lot more Law of Demeter violations.

It's still a God class indeed. Even if EMS was moved into the DOS class, and GDB specific methods were moved to the GDB namespace, there is more to do. Right now, it's partionned into loigcal part in order to make sense of it. But, it's still highly coupled.

There is more to read on one screen. Way less DMA noise getting in the way when reading the Machine class, for example. 
But there still a lot to do in terms of IoC.

But the part about

src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
JorisVanEijden: The way it was, a _reference_ to an `IVideoCard` was injected. As long as that reference pointed to nothing, nothing happened (due to the `?`). Once there _was_ something implementing that interface there, it had it's UpdateScreen method called.

The way you do it now the timer has to depend on a Machine which has to depend on a VideoSubSystem which has to depend on a VgaCard. All of which now have to exist and point to something, even though none of it is required by the Timer here. That only needs a reference to an `IVideoCard` to call UpdateScreen() on if it's filled.

The end goal is to have all the constructors parameters filled with only exactly the nbterfaces they themselves need.
And then have those be provided by StrongInject right?

src/Spice86/Program.cs
JorisVanEijden: Yes doing IConfiguration or full-on Options pattern heree and now would not make sense.
But moving the Configuration creation to _before_ the container initialization does feel like something that belongs in this PR.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
maximilien-noal: Done.

src/Spice86/Startup.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
maximilien-noal: >  And then have those be provided by StrongInject right?

Yep. Although I'm not a fan of calling  the Resolve method of the container in our code.
This seems to be common with Source Generator based DI frameworks....

Putting everything (both emulator, UI, and other services) in one IContanier will be a nightmare to mock.

OTOH, using several containers we will use the Resolve

The code should be written as if This is what is done by user code in an ASP .NET Core webapp thanks to the reflection based magic of the Microsoft DI nugget package with which the user code first registers services (for each there is an interface and a set lifetime), and when a HTTP request hits a MVC controller, the Generic Host instantiates it itself. You don't have to Resolve services.

Maybe we could depend on the IContainer interface for the emulator container / UI container (depending on the layer) on the root object (for the UI or console, this is Program.cs).

But what is the root object for the emulator layer ? The ProgramExecutor ?
But the SoundBlaster for example will not be instantiated by the ProgramExecutor.
As of now, it is instantiated by the Machine. Should we resolve the container service there ?
But then the Machine ctor will depend on the IContainer interface, and then at this stage we might put the emulator IContainer as a dependcy for every class in the emulator assembly.
Or maybe for sound, have a SoundContainer... ?!

Yeah, I'm not a fan of the DI framework being apparent in the code. It should be transparent. 

AvaloniaUI's host both does not offer such a feature, and is incompatible with the Generic Host...
Plus we instantiate the emulator ourselves, we don't react to HTTP methods...

BTW, I corrected the Timer. It now depends directly on the IVideoCard.

JorisVanEijden: The trick with a DI framework is to never `new` anything that has an interface (which everything that has functionality should have).

`ProgramExecutor` needs a `Machine`? Inject an `IMachine` into its constructor.
`Machine` needs a `MachineCreationOptions`? Inject an `IMachineCreationOptions` into its constructor.
`MachineCreationOptions` needs an `IGui`, an `ILoggerService`, a `CounterConfigurator`, an `ExecutionFlowRecorder`, and a `Configuration`? Inject their interfaces into its constructor.
etc.

This makes the DI completely transparant. But it also requires a throrough understanding of all the (inter)dependencies in the code base, and elimination of cyclical dependencies, etc.

I feel this a bridge too far for this project at this stage.

Like I said before, we are currently using the `Machine` class as our dependency container.
Almost all dependencies are instantiated there, the order and requirements of the instantiations is taken care of there, the configuration of implementations, basically everything a dependency injection framework does is done there.
We also use it as a Service Locator. It is passed into almost all classes to provide a way for those classes to Get their dependencies from the Machine.

Rather than hiding this functionality deeper into "subsystems" I'd suggest splitting the Machine class in two : 1 "Machine" and 1 "ServiceContainer".
Isolating all the "Dependency Management/ Service Container" functionality makes it possible to maybe later on move it into a DI framework.
Or we may choose to keep our own homegrown ServiceContainer (this is now hidden inside the Machine) and build on it and improve it.

All this is not essential to the testability of the code. Interfaces are. Replacing every concrete (non-data-object) class with an interface in the constructor parameters will help immensly in making the code unit testable. And since the three most common parameters are `Machine`, `Configuration` and `ILoggerService` 

One approach would be to slowly replace the Machine(in its role of ServiceContainer) injection in the "leaves" of our codebase with the interfaces that the code actually uses. That way, usage of the Machine will move more and more to the "top" where it can eventually be replaced by something like StrongInject.

src/Spice86/Program.cs
maximilien-noal: I moved the Configuration creation before the line that created the Container.
I pass it the configuration on initialization, and use its Configuration property to call the Startup.StartApp.

But I've yet to use that property everywhere in the UI layer, as it is only available in Program.cs since topLevelContainer is a local variable.
To avoid a warning in Program.cs saying "captured variable disposed in the outer scope" I don't dispose of the container in Program.cs anymore...

src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
kevinferrare: I feel you are right.
But there is one thing I think we need to consider this is that devices lifecycle is naturally bound to the Machine.
You encouter HLT -> you dont only stop execution, you also reset all your devices.
Maybe execution part of the machine could be in another class?

PR #358, opened 2023-06-29
Feature: Create/Remove Directory. Feature: GetCurrentDirectory. Featue: Select and Get Default Drive. Fix: '..' and '.' support.

### LINKED ISSUE #326: Betrayal at Krondor bug related to file/directory handling.
**Describe the bug**
When you start Betrayal at Krondor without theree being a "GAMES" directory in the same directory as the krondor.exe file, then graphics corruption occurs after starting chapter I.

**To Reproduce**
1. Make sure there is no GAMES directory in the directory where krondor.exe is
2. Start the game
3. Start Chapter I
4. See corrupted graphics

**Expected behavior**
No graphics corruption.

**Screenshots**
Without GAMES folder:
image
With GAMES folder:
image

**Desktop (please complete the following information):**
 - OS: both windows and linux
 - Version: both master and new_enhanced_videocard branch

**Additional context**
This bug does not appear on Dosbox-X.
I suspect it has something to do with file and or directory handling in DOS. 
The game appear sto do some funky creating, searching and deleting of files/directories during the start of a Chapter.

maximilien-noal:
### Description of Changes

Introduces the DOS INT21H ability to create a folder (including recursively).

Introduces the DOS INT21H ability to remove a folder (but not recursively).

Introduces the DOS INT21H ability to get the current DOS directory.

Introduces the DOS INT21H ability to set the current folder to relative folders '.' and '..' (which are special DOS 'folders representing the current folder and the parent folder, respectively).

Fixes the methods for setting / getting the default drive.

The Current Directory is also now different depending on the drive. 

### Rationale behind Changes

Fixes #326 

### Suggested Testing Steps

Already tested (including savegame support):
- Betrayal at Krondor (both with and without a pre-created GAMES folder)
- Dune
- Dune II
- Duke Nukem II


src/Spice86.Shared/Utils/ConvertUtils.cs
```diff
@@ -362,7 +362,7 @@ public static class ConvertUtils {
     /// <param name="path">The path string to modify.</param>
     /// <returns>A new string with all backslashes replaced with forward slashes.</returns>
     public static string ToSlashPath(string path) {
-        return path.Replace('\\', '/').Replace("//", "/");
+        return path.Replace('\\', Path.DirectorySeparatorChar).Replace("//", Path.DirectorySeparatorChar.ToString());
```
kevinferrare, line 365: Then maybe function name needs to change because on windows it will be \
Or maybe it is better to keep / since it would make emulator behavior the same for any host OS?
Not sure what is best.
Why you needed to change this by the way.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -717,4 +762,43 @@ public class DosFileManager {
...
+            return FileNotFoundError(directory);
+        }
+
+        DirectoryInfo? parentFolder = Directory.GetParent(fullPath);
+
+        if (parentFolder?.Exists is false or null) {
+            return FileNotFoundError(directory);
+        }
+
+        if (parentFolder.GetDirectories().Select(x => x.Name)
```
kevinferrare, line 786: As far as I understand, this LINQ expression lists the name of the files and folders of the parent. Maybe it would be good to move it to a method with a name that reflects that (because I had to think about what it does for more that an few seconds :) )

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -717,4 +762,43 @@ public class DosFileManager {
...
+            .Concat(parentFolder.GetFiles().Select(x => x.Name))
+            .Any(x => string.Equals(x, directory, StringComparison.InvariantCultureIgnoreCase))) {
```
kevinferrare, line 788: directory can contain more than the directory name isnt it? In this case I am not sure to understand how it is going to be equal to files / folders names

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -673,36 +707,47 @@ public class DosFileManager {
...
 
         return ConvertUtils.ToSlashPath(fileName);
     }
 
+    /// <summary>
+    /// All the possible DOS drive letters
+    /// </summary>
+    public string[] DriveLetters { get; } = {"A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z" };
+    
+    private bool IsDosPathRooted(string fileName) => fileName.Length >= 2 && DriveLetters.Contains(fileName[0].ToString().ToUpperInvariant()) && fileName[1] == ':';
```
kevinferrare, line 749: Really makes me think all the files and folders conversion / browsing stuff (mapping between host and emulated files) should be in a separate class considering how complex this is becoming

src/Spice86.Shared/Utils/ConvertUtils.cs
maximilien-noal: I reverted this cosmetic change which had no real value.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: It lists folders first, files next, takes their names, and determines if any equals with the new directory name.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: This is on-going.

One thing missing is 8.3 file names.

There is another class wanting to get out of the DOS file manager, indeed.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: Moved to another method.

Sorry about that, I'm a LINQ addict.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -717,4 +762,43 @@
...
+            return FileNotFoundError(directory);
+        }
+
+        DirectoryInfo? parentFolder = Directory.GetParent(fullPath);
+
+        if (parentFolder?.Exists is false or null) {
+            return FileNotFoundError(directory);
+        }
+
+        if (parentFolder.GetDirectories().Select(x => x.Name)
```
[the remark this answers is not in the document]
maximilien-noal: CodeQL you're drunk. Go home.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: DosFIleResolver class created, along with interface.

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
```diff
@@ -0,0 +1,172 @@
...
+    private readonly ILoggerService _loggerService;
+
+    /// <summary>
+    /// Initializes a new instance.
+    /// </summary>
+    /// <param name="loggerService">The logger service implementation.</param>
+    public DosFilePathResolver(ILoggerService loggerService) => _loggerService = loggerService;
+    
+    /// <inheritdoc />
+    public string GetHostParentDirectory(string path) => Directory.GetParent(path)?.FullName ?? path;
```
kevinferrare, line 22: Not sure this has something to do with Host. Maybe just "GetParentDirectory"? Also, could it be used in places where Directory.GetParent is used?

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
```diff
@@ -0,0 +1,172 @@
...
+    private static string? GetDirectoryCaseSensitive(string? directory) {
+        if (string.IsNullOrWhiteSpace(directory)) {
+            return null;
+        }
+        DirectoryInfo directoryInfo = new(directory);
+        if (directoryInfo.Exists) {
+            return directory;
+        }
+
+        if (directoryInfo.Parent == null) {
```
kevinferrare, line 55: Why not storing parent in a local variable since it is going to be reused twice after that?

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
```diff
@@ -0,0 +1,172 @@
...
+    /// <inheritdoc />
+    public string? GetActualCaseForFileName(string caseInsensitivePath) {
+        string? directory = Path.GetDirectoryName(caseInsensitivePath);
+        string? directoryCaseSensitive = GetDirectoryCaseSensitive(directory);
+        if (string.IsNullOrWhiteSpace(directoryCaseSensitive) || Directory.Exists(directoryCaseSensitive) == false) {
+            return null;
+        }
+        string realFileName = "";
+        string[] array = Directory.GetFiles(directoryCaseSensitive);
+        foreach (string file in array) {
```
kevinferrare, line 33: Looks like this is already done in GetDirectoryCaseSensitive isnt it?

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
```diff
@@ -0,0 +1,172 @@
...
+                realFileName = file;
+            }
+        }
+        if (string.IsNullOrWhiteSpace(realFileName) || File.Exists(realFileName) == false) {
+            return null;
+        }
+        return realFileName;
+    }
+
+    private static string? GetDirectoryCaseSensitive(string? directory) {
```
kevinferrare, line 46: From what I understand this method gets the case sensitive name of the leaf directory in the path if leaf parent exists. Not sure the name is correct.

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
maximilien-noal: Renamed, and used everywhere it made sense.

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
maximilien-noal: It's a recursive call to itself, which can return null if:

if (string.IsNullOrWhiteSpace(directory)) {
            return null;
        }
        DirectoryInfo directoryInfo = new(directory);
        if (directoryInfo.Exists) {
            return directory;
        }

        if (directoryInfo.Parent == null) {
            return null;
        }

        string? parent = GetDirectoryCaseSensitive(directoryInfo.Parent.FullName);
        if (parent == null) {
            return null;
        }

        The compiler forces checking for null if you mark anything nullable.

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
```diff
@@ -0,0 +1,172 @@
...
+        DirectoryInfo directoryInfo = new(directory);
+        if (directoryInfo.Exists) {
+            return directory;
+        }
+
+        if (directoryInfo.Parent == null) {
+            return null;
+        }
+
+        string? parent = GetDirectoryCaseSensitive(directoryInfo.Parent.FullName);
```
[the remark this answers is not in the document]
maximilien-noal: It's a recursive call to itself, which can return null if:

```

if (string.IsNullOrWhiteSpace(directory)) {
            return null;
        }
[...]

        if (directoryInfo.Parent == null) {
            return null;
        }

        string? parent = GetDirectoryCaseSensitive(directoryInfo.Parent.FullName);
        if (parent == null) {
            return null;
        }
```

Furthermore, the compiler always considers it can be null if you mark anything nullable.

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
maximilien-noal: Renamed method.

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
maximilien-noal: I don't think so. Here, we list files in the folder returned by GetDirectoCaseSensitive.

I renamed the method to TryGetHostFullNameForFile and removed a useless check.

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
maximilien-noal: Sorry, I don't get it ?

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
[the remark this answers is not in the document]
maximilien-noal: I  refactored the parameter so it is now a string, instead of a nullable string.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        CurrentHostDirectory = Path.GetFullPath(Path.Combine(CurrentHostDirectory, newCurrentDir));
+        return DosFileOperationResult.NoValue();
+    }
+
+    /// <inheritdoc/>
+    public void SetDiskParameters(char currentDrive, string newCurrentDir, Dictionary<char, MountedFolder> driveMap) {
+        DriveMap = driveMap;
+        CurrentDrive = currentDrive;
+        SetCurrentDir(newCurrentDir);
```
JorisVanEijden, line 71: `SetCurrentDir()` expects a DOS path, not a host path.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -146,26 +149,33 @@ public class DosFileManager {
...
-                        RecurseSubdirectories = false
-                    }).ToList();
-                _matchingFilesIterator = matchingPaths.GetEnumerator();
-                return FindNextMatchingFile();
-            } catch (IOException e) {
-                e.Demystify();
-                if (_loggerService.IsEnabled(Serilog.Events.LogEventLevel.Error)) {
-                    _loggerService.Error(e, "Error while walking path {CurrentMatchingFileSearchFolder} or getting attributes", _currentMatchingFileSearchFolder);
-                }
+        string hostSearchSpec = _dosPathResolver.PrefixWithHostDirectory( fileSpec);
```
kevinferrare, line 152: leading space before filespec (formatting)

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -437,93 +433,62 @@ public class DosFileManager {
...
+    /// <ul>
+    /// <li>Prefix either the current folder or the drive folder.</li>
+    /// <li>Replace backslashes with slashes</li>
+    /// <li>Find case sensitive matches for every path item (since DOS is case insensitive but some OS are not)</li>
+    /// </ul>
+    /// </summary>
+    /// <param name="dosFileName">The file name to convert.</param>
+    /// <param name="forCreation">if true will try to find case sensitive match for only the parent of the file</param>
+    /// <returns>the file name in the host file system, or null if nothing was found.</returns>
+    internal string? ToHostCaseSensitiveFileName(string dosFileName, bool forCreation) => _dosPathResolver.ToHostCaseSensitiveFullName( dosFileName, forCreation);
```
kevinferrare, line 476: forCreation -> Although I think it was me who wrote that, I am not sure it is the best name, maybe convertParentOnly would be more representative?

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -602,119 +556,57 @@ public class DosFileManager {
...
-    private string? ToCaseSensitiveFileName(string? caseInsensitivePath) {
-        if (string.IsNullOrWhiteSpace(caseInsensitivePath)) {
-            return null;
+    /// <summary>
+    /// Creates a directory on disk.
+    /// </summary>
+    /// <param name="directory">The directory name to create</param>
+    /// <returns></returns>
+    /// <returns>A <see cref="DosFileOperationResult"/> with details about the result of the operation.</returns>
+    public DosFileOperationResult CreateDirectory(string directory) {
```
kevinferrare, line 582: directory -> dosDirectory?

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -602,119 +556,57 @@ public class DosFileManager {
...
+        string hostPath = _dosPathResolver.PrefixWithHostDirectory(directory);
+
+        string? fullPath = _dosPathResolver.ToHostCaseSensitiveFullName(directory, true);
```
kevinferrare, line 585: Not sure to understand why converting the param in input and not hostpath

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -602,119 +556,57 @@ public class DosFileManager {
...
-            string? fileNameOnFileSystem = GetActualCaseForFileName(caseInsensitivePath);
-            if (string.IsNullOrWhiteSpace(fileNameOnFileSystem) == false) {
-                return fileNameOnFileSystem;
-            }
-            Regex fileToProcessRegex = FileSpecToRegex(Path.GetFileName(fileToProcess));
-            if (Directory.Exists(parent)) {
-                return Array.Find(Directory
-                    .GetFiles(parent), x => fileToProcessRegex.IsMatch(x));
-            }
+            Directory.CreateDirectory(hostPath);
```
kevinferrare, line 601: This may fail because hostPath is not case sensitive right?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,236 @@
...
+using Spice86.Shared.Utils;
+
+/// <inheritdoc />
+public class DosPathResolver : IDosPathResolver {
+
+    /// <inheritdoc />
+    public IDictionary<char, MountedFolder> DriveMap { get; private set; } = new Dictionary<char, MountedFolder>();
+
+    /// <inheritdoc />
+    public char CurrentDrive { get; private set; }
```
JorisVanEijden, line 20: Non-nullable property is not set after invoking constructor

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,236 @@
...
+
+    /// <inheritdoc />
+    public string CurrentHostDirectory {
```
JorisVanEijden, line 23: Non-nullable property is not set after invoking constructor

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: This is not reported by the compiler, as char has a default non-null value.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: This is not reported by the compiler, as this is a calculated property.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,236 @@
...
+                return DosFileOperationResult.NoValue();
+            }
+        }
+
+        if (newCurrentDir == ".") {
+            return DosFileOperationResult.NoValue();
+        }
+
+        if (newCurrentDir == "..") {
+            CurrentHostDirectory = GetFullNameForParentDirectory(CurrentHostDirectory) ?? CurrentHostDirectory;
```
JorisVanEijden, line 54: `GetFullNameForParentDirectory` never returns `null` so the `??` does nothing

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,236 @@
...
+            return DosFileOperationResult.NoValue();
+        }
+
+        while (newCurrentDir.StartsWith("..\\")) {
+            newCurrentDir = newCurrentDir[3..];
+            CurrentHostDirectory = GetFullNameForParentDirectory(CurrentHostDirectory) ?? CurrentHostDirectory;
```
JorisVanEijden, line 60: GetFullNameForParentDirectory never returns null so the ?? does nothing

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: renamed to convertParentOnly.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: I moved hostPath closer to where it's used.

hostPath is for creation of the folder, fullPath is for verification if we can do it.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: Renamed to dosDirectory.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: This is why IsThereAnyDirectoryOrFileWithTheSameName checks if we can create the folder first.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -30,32 +28,37 @@
...
     static DosFileManager() {
-        _fileOpenMode.Add(0x00, "r");
-        _fileOpenMode.Add(0x01, "w");
-        _fileOpenMode.Add(0x02, "rw");
+        FileOpenMode.Add(0x00, "r");
+        FileOpenMode.Add(0x01, "w");
+        FileOpenMode.Add(0x02, "rw");
     }
 
+    private IDosPathResolver _dosPathResolver;
```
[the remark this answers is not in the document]
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
JorisVanEijden: I meant that you should make it nullable if you are not going to give it a default value. That way code that uses the interface can see  that this property may not have a valid value yet.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
JorisVanEijden: I meant that you should make it nullable if you are not going to give it a default value. That way code that uses the interface can see that this property may not have a valid value yet.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+
+    private bool StartsWithDosDrive(string path) => 
+        path.Length >= 2 &&
+        DriveLetters.Contains(path[0].ToString().ToUpperInvariant()) &&
+        path[1] == ':';
+
+    private bool IsPathRooted(string path) => 
+        path.Length >= 3 &&
+        StartsWithDosDrive(path) &&
+        (path[2] == '\\' || path[2] == '/');
```
JorisVanEijden, line 228: I didn't think DOS paths could have a forward slash?

Also, I don't think `C=\` should be considered "rooted"

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: C:\ is considered by System.IO.Path to be rooted (on Windows) in the .NET BCL.
The same should be true in DOS, as Windows is retro compatible.

This can deal with forward slashes, since the caller can give a parameter that has been already internally converted to forward slashes.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+    private static Regex FileSpecToRegex(string fileSpec) {
+        string regex = fileSpec.ToLowerInvariant();
+        regex = regex.Replace(".", "[.]");
+        regex = regex.Replace("?", ".");
+        regex = regex.Replace("*", ".*");
+        return new Regex(regex);
+    }
+    
+    /// <inheritdoc />
+    public string? ToHostCaseSensitiveFullName(string dosFileName, bool convertParentOnly) {
```
JorisVanEijden, line 180: Could we have two separate methods, rather than a boolean flag?
a `ToHostCaseSensitiveFullName` and `ToHostCaseSensitiveParentName` or something?

src/Spice86.Core/Emulator/OperatingSystem/IDosPathResolver.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <param name="convertParentOnly">if true, it will try to find the case sensitive match for only the parent of the path</param>
+    /// <returns>A string containing the full file path in the host file system, or <c>null</c> if nothing was found.</returns>
+    string? ToHostCaseSensitiveFullName(string dosPath, bool convertParentOnly);
+
+    /// <summary>
+    /// Returns the full host file path, including casing.
+    /// </summary>
+    /// <param name="dosFilePath">The DOS file path.</param>
+    /// <returns>A string containing the host file path, or <c>null</c> if not found.</returns>
+    string? TryGetFullHostFileName(string dosFilePath);
```
JorisVanEijden, line 47: What is the difference between these two methods?
Or, how would a user of this interface know which one to use?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+        CurrentDrive = currentDrive;
+        SetCurrentDir(dosPath);
+    }
+
+    /// <inheritdoc />
+    public string GetFullNameForParentDirectory(string path) => Directory.GetParent(path)?.FullName ?? path;
+
+    /// <inheritdoc />
+    public string? TryGetFullHostFileName(string dosFilePath) {
+        string? directory = Path.GetDirectoryName(dosFilePath);
```
JorisVanEijden, line 78: This takes a DOS filepath, but then uses it to search the host filesystem?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+
+    /// <inheritdoc />
+    public IDictionary<char, MountedFolder> DriveMap { get; private set; } = new Dictionary<char, MountedFolder>();
+
+    /// <inheritdoc />
+    public char CurrentDrive { get; private set; }
+
+    /// <inheritdoc />
+    public string CurrentHostDirectory {
+        get => DriveMap[CurrentDrive].CurrentFolder;
```
JorisVanEijden, line 23: On windows this returns a path with backslashes while other functions return paths with forward slashes.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+
+    /// <inheritdoc />
+    public string GetHostRelativePathToCurrentDirectory(string path) => Path.GetRelativePath(CurrentHostDirectory, path);
+
+    /// <inheritdoc/>
+    public DosFileOperationResult SetCurrentDir(string dosPath) {
+        if (IsPathRooted(dosPath)) {
+            string? newCurrentDirectory = ToHostCaseSensitiveFullName(dosPath, false);
+            if (!string.IsNullOrWhiteSpace(newCurrentDirectory)) {
+                CurrentHostDirectory = newCurrentDirectory;
```
JorisVanEijden, line 43: If the drive in the dosPath is different from the currentDrive, then this sets the path to the wrong drive.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -602,119 +557,56 @@ public class DosFileManager {
...
+        if (!Directory.Exists(parentFolder)) {
+            CreateDirectory(_dosPathResolver.GetHostRelativePathToCurrentDirectory(parentFolder));
         }
 
-        string? parent = ToCaseSensitiveFileName(parentDir);
-        if (parent == null) {
-            // End of recursion, root reached
-            return null;
+        if (_dosPathResolver.IsThereAnyDirectoryOrFileWithTheSameName(dosDirectory, new DirectoryInfo(parentFolder))) {
+            return FileNotFoundError(dosDirectory);
```
JorisVanEijden, line 596: I suspect this is not the correct error.

JorisVanEijden: This is also susceptible to race conditions, where the file or directory is created between this check and the directory creation.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
JorisVanEijden: I see that Dosbox-x allows both forward and backward slashes. I'm not sure what DOS on real hardware does.

What I meant by `C=\` not being considered rooted, is because `=` is an equals sign, not a colon.
`CC/foo` is also considered "rooted" by this method. I think you should check for a `:` no?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+            path = Path.Combine(CurrentHostDirectory, path);
+        }
+
+        return ConvertUtils.ToSlashPath(path);
+    }
+
+    /// <summary>
+    /// All the possible DOS drive letters
+    /// </summary>
+    private static string[] DriveLetters { get; } = { "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z" };
```
JorisVanEijden, line 218: Suggestion:
```suggestion
    private static IEnumerable<char> DriveLetters => "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+
+    private bool StartsWithDosDrive(string path) => 
+        path.Length >= 2 &&
+        DriveLetters.Contains(path[0].ToString().ToUpperInvariant()) &&
```
JorisVanEijden, line 222: Suggestion:
```suggestion
        DriveLetters.Contains(char.ToUpperInvariant(path[0])) &&
```

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+    public DosFileOperationResult SetCurrentDir(string dosPath) {
+        if (IsPathRooted(dosPath)) {
+            string? newCurrentDirectory = ToHostCaseSensitiveFullName(dosPath, false);
+            if (!string.IsNullOrWhiteSpace(newCurrentDirectory)) {
+                CurrentHostDirectory = newCurrentDirectory;
+                return DosFileOperationResult.NoValue();
+            }
+        }
+
+        if (dosPath == ".") {
```
JorisVanEijden, line 48: what if the path is `.\` ?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+        }
+        string? directoryCaseSensitive = TryGetFullNameOnDiskOfParentDirectory(directory);
+        if (string.IsNullOrWhiteSpace(directoryCaseSensitive) || !Directory.Exists(directoryCaseSensitive)) {
+            return null;
+        }
+        string hostFileName = "";
+        string[] array = Directory.GetFiles(directoryCaseSensitive);
+        foreach (string file in array) {
+            string fileToUpper = file.ToUpperInvariant();
+            string searchedFile = dosFilePath.ToUpperInvariant();
```
JorisVanEijden, line 90: Suggestion: move this outside the loop

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: I consider null to be a code smell. Null questions in any code base make it less readable. I'd really like to avoid it if possible.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: I consider null to be a code smell. Null questions in any code base make it less readable. I'd really like to avoid it if possible.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: GetDirectoryName doesn't work with the host file system. It just parses the input.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: This is taken into account now.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: TOCTOU problem, yes. But the host file system is not in our control, isn't it ?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: I believed to have fixed it in the latest version.

src/Spice86.Core/Emulator/OperatingSystem/IDosPathResolver.cs
maximilien-noal: This is still the most confusing part remaining.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: I'd like to remove that boolean too.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: That's what StartsWithDosDrive does (checking for ':') and it's called by IsPathRooted.

Can't see how it would consider C=\ to be valid.

Speaking of, we miss DOS invalid characters and invalid names checks.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
JorisVanEijden: Then make sure to initialize these properties from the constructor.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
JorisVanEijden: Sorry, misread the dosdrive part. I thouhg it only checked if the drive letter was mounted.

Invalid character and name checks should probably getr a centralized place somewhere once everything is settled a bit.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
JorisVanEijden: I meant in the foreach a bit later in the code.
If I pass it a DOS path like "D:TOOLS" it will read all the files on my host D drive.

If you look at how it is used you have:
- `Spice86.Core.Emulator.OperatingSystem.DosFileManager.OpenFileInternal()` which passes it a host path.
- `Spice86.Core.Emulator.OperatingSystem.DosPathResolver.ToCaseSensitiveFileName` which passes it a DOS path

They can't both be right.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
JorisVanEijden: It is not. And I think the check is only to see if another file, with a different case, on a case sensitive filesystem, exists. To get an error that the CreateDirectory would not.
So the check is good. It's just the type of error I'm not sure about. Does DOS really throw a FileNotFound if a directory already exists?

src/Spice86.Core/Emulator/OperatingSystem/IDosPathResolver.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <returns>A <see cref="DosFileOperationResult"/> that details the result of the operation.</returns>
+    DosFileOperationResult SetCurrentDir(string newCurrentDir);
+
+    /// <summary>
+    /// Sets the current DOS folder, and the map between DOS drive letters and host folders paths.
+    /// </summary>
+    /// <param name="currentDrive">The current DOS drive letter.</param>
+    /// <param name="newCurrentDir">The new host folder to use as the current DOS folder.</param>
+    /// <param name="driveMap">The map between DOS drive letters and host folders paths.</param>
+    void SetDiskParameters(char currentDrive, string newCurrentDir, Dictionary<char, MountedFolder> driveMap);
```
JorisVanEijden, line 87:
```suggestion
    void SetDiskParameters(char currentDrive, string newCurrentDir, IDictionary<char, MountedFolder> driveMap);
```

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,232 @@
...
+    private static IEnumerable<char> DriveLetters => "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+    private bool StartsWithDosDrive(string path) => 
+        path.Length >= 2 &&
+        DriveLetters.Contains(char.ToUpperInvariant(path[0])) &&
+        path[1] == ':';
+
+    private bool IsPathRooted(string path) => 
+        path.Length >= 3 &&
+        StartsWithDosDrive(path) &&
```
JorisVanEijden, line 224: But, paths that start with a slash are also rooted aren't they?
I mean, if I'm in `d:\data\docs` and type `cd \Tools` then I'm in `d:\tools`

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/OperatingSystem/IDosPathResolver.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/IDosPathResolver.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Paths that start with '\' are now considered to be rooted.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: It is the correct error code. In case of error, it either returns 0x2 (file not found) or 0x5 (access denied, in case of network access, starting with MS-DOS 3.3)

See:
https://www.gladir.com/LEXIQUE/INTR/int21f39.htm
also:
https://marc.info/?l=wine-patches&m=104897290211098&w=2 (msdos/int21.c in wine)
https://gitlab.winehq.org/szanni/wine/-/blob/wine-980517/msdos/int21.c

maximilien-noal:
PR tested again with Dune II, Dune, Krondor, and Duke Nukem II successfully.

The DOSPathResolver is really simplified now.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Property set in the ctor.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -513,18 +534,34 @@
...
+    /// <returns>
+    /// DS:DI = ASCIZ pathname containing the current DOS directory. <br/>
+    /// CF is clear if sucessful. <br/>
+    /// CF is set on error. Possible error code in AX: 0xF (InvalidDrive).
+    /// </returns>
+    /// <param name="calledFromVm">Whether the method was called by the emulator.</param>
     public void GetCurrentDirectory(bool calledFromVm) {
-        SetCarryFlag(false, calledFromVm);
+        uint responseAddress = MemoryUtils.ToPhysicalAddress(_state.DS, _state.SI);
+        DosFileOperationResult result = _dosFileManager.GetCurrentDir(_state.DL, out string currentDir);
```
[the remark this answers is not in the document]
maximilien-noal: Fixed.

src/Spice86.Tests/Dos/DosFileManagerTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        string tempDir = Path.GetTempPath();
+        if(!Directory.Exists($"{tempDir}/TEST")) {
+            Directory.CreateDirectory($"{ tempDir}/TEST");
+        }
+        Configuration configuration = new Configuration() {
+            DumpDataOnExit = false,
+            CDrive = tempDir
+        };
+        IMemoryDevice ram = new Ram(A20Gate.EndOfHighMemoryArea);
+        IMemory memory = new Memory(ram, configuration);
```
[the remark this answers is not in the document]
maximilien-noal: Easier for debugging.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// Gets the DOS drive letter.
+    /// </summary>
+    public char DriveLetter { get; init; }
+
+    /// <summary>
+    /// The full host path to the mounted folder. This path serves as the root of the DOS drive.
+    /// </summary>
+    public string MountPoint { get; init; }
```
kevinferrare, line 26: Would be clearer if every path had a host / dos prefix to get better what it is at glance, especially when it is used

src/Spice86.Core/Emulator/OperatingSystem/IDosPathResolver.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// <summary>
+    /// The current DOS drive in use.
+    /// </summary>
+    char CurrentDrive { get; }
+
+    /// <summary>
+    /// The full host path to the folder used by DOS as the current folder.
+    /// </summary>
+    string CurrentHostDirectory { get; }
```
JorisVanEijden, line 24: These three properties are never used outside the DosPathResolver itself. Is there a usecase for them to be accessible through the interface?

maximilien-noal: I'd like to extend the debug window in the UI a lot, and show those properties, eventually.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+    /// <param name="loggerService">The logger service implementation.</param>
+    /// <param name="configuration">The emulator configuration.</param>
+    public DosPathResolver(ILoggerService loggerService, Configuration configuration) {
+        _loggerService = loggerService;
+        DriveMap = InitializeDriveMap(configuration);
+        CurrentDrive = 'C';
+        SetCurrentDir(@"C:\");
+    }
+
+    private static string GetExeParentFolder(Configuration configuration) {
```
JorisVanEijden, line 56: Why not pass in a `string?` and call the method `GetParentFolder`?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+        DirectoryInfo? parentDir = Directory.GetParent(exe);
+        // Must be in the current directory
+        parentDir ??= new DirectoryInfo(Environment.CurrentDirectory);
+
+        string parent = Path.GetFullPath(parentDir.FullName);
+        return ConvertUtils.ToSlashFolderPath(parent);
+    }
+
+    private IDictionary<char, MountedFolder> InitializeDriveMap(Configuration configuration) {
+        string parentFolder = GetExeParentFolder(configuration);
```
JorisVanEijden, line 71: Move this inside the `if` so it only gets executed when there is no cDrive in the configuration.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
```diff
@@ -0,0 +1,37 @@
...
+/// <summary>
+/// Represents a host folder used as a drive by DOS.
+/// </summary>
+public class MountedFolder {
+    /// <summary>
+    /// Initializes a new instance.
+    /// </summary>
+    /// <param name="driveLetter">The DOS driver letter.</param>
+    /// <param name="fullName">The full host path to the folder to be used as the DOS drive root.</param>
+    public MountedFolder(char driveLetter, string fullName) {
```
JorisVanEijden, line 12:
```suggestion
    public MountedFolder(char driveLetter, string mountPoint) {
```

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// The current directory in use on the drive. Relative to the <see cref="MountPoint"/>
+    /// </summary>
+    public string CurrentDirectory { get; set; }
+
+    /// <summary>
+    /// The full host path. Combined from <see cref="MountPoint"/> and <see cref="CurrentDirectory"/>
+    /// </summary>
+    public string FullName => Path.Combine(MountPoint, CurrentDirectory);
```
JorisVanEijden, line 36:
```suggestion
    public string FullHostPath => Path.Combine(MountPoint, CurrentDirectory);
```

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,235 @@
...
+        Dictionary<char, MountedFolder> driveMap = new();
+        string? cDrive = configuration.CDrive;
+        if (string.IsNullOrWhiteSpace(cDrive)) {
+            cDrive = parentFolder;
+        }
+        cDrive = ConvertUtils.ToSlashFolderPath(cDrive);
+        driveMap.Add('C', new MountedFolder('C', cDrive));
```
JorisVanEijden, line 78: What if I don't want a C drive?

maximilien-noal: Currently, there is always a CDrive. It either comes from the CDrive property, or it comes from parsing the Configuration.ExePath

We don't support right now not having a C Drive.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Currently, there is always a CDrive. It either comes from the CDrive property, or it comes from parsing the Configuration.ExePath

We don't support right now not having a C Drive.

src/Spice86.Core/Emulator/OperatingSystem/DosFilePathResolver.cs
maximilien-noal: Set outdated code comment as resolved. Please re-open the conversation if the issue still applies.

maximilien-noal:
Just rebased on master. Will continue on this tonight or tomorrow.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -517,18 +530,35 @@ public class DosInt21Handler : InterruptHandler {
...
         if (_loggerService.IsEnabled(LogEventLevel.Verbose)) {
-            _loggerService.Verbose("GET CURRENT DIRECTORY {ResponseAddress}",
-                ConvertUtils.ToSegmentedAddressRepresentation(_state.DS, _state.SI));
+            _loggerService.Verbose("GET CURRENT DIRECTORY {ResponseAddress}: {CurrentDpsDirectory}",
+                ConvertUtils.ToSegmentedAddressRepresentation(_state.DS, _state.SI), currentDir);
         }
-        uint responseAddress = MemoryUtils.ToPhysicalAddress(_state.DS, _state.SI);
-        // Fake that we are always at the root of the drive (empty String)
-        _memory.UInt8[responseAddress] = 0;
+        _memory.LoadData(responseAddress, Encoding.ASCII.GetBytes(currentDir).Concat(Encoding.ASCII.GetBytes("0")).ToArray());
```
kevinferrare, line 554: Why not using Memory.SetZeroTerminatedString?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,264 @@
...
+                RecurseSubdirectories = true,
+            })
+            .Select(x => (FileSystemInfo)x)
+            .Concat(
+            hostDirInfo.EnumerateFiles("*",
+            new EnumerationOptions() {
+                RecurseSubdirectories = true,
+            }).Select(x => x))
+            .FirstOrDefault(x =>
+                x is FileInfo ?
```
kevinferrare, line 195: Is it possible to move the lambda to a separate method? The ternary is long and makes my head hurt :)

maximilien-noal: Done.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,264 @@
...
+        string? relativeHostPath = hostDirInfo
+            .EnumerateDirectories("*", new EnumerationOptions() {
+                RecurseSubdirectories = true,
+            })
+            .Select(x => (FileSystemInfo)x)
+            .Concat(
+            hostDirInfo.EnumerateFiles("*",
+            new EnumerationOptions() {
+                RecurseSubdirectories = true,
+            }).Select(x => x))
```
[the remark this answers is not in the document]
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,264 @@
...
+    /// <summary>
+    /// Create a relative path from the current host directory to another. Paths will be resolved before calculating the difference.
+    /// Default path comparison for the active platform will be used (OrdinalIgnoreCase for Windows or Mac, Ordinal for Unix).
+    /// </summary>
+    /// <param name="hostPath">The destination path.</param>
+    /// <returns>A string containing the relative host path, or <paramref name="hostPath"/> if the paths don't share the same root.</returns>
+    public string GetRelativeHostPathToCurrentDirectory(string hostPath) => Path.GetRelativePath(CurrentHostDirectory, hostPath);
+
+    private static bool IsWithinMountPoint(string hostFullPath, MountedFolder mountedFolder) =>
+        hostFullPath.StartsWith(mountedFolder.MountPoint);
```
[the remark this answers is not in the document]
maximilien-noal: Normalization done.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,272 @@
...
+        string relativePath = fileOrDirInfo.FullName[HostPrefix.Length..];
+        if (fileOrDirInfo is FileInfo) {
+            return string.Equals(ConvertUtils.ToSlashPath(relativePath),
+                ConvertUtils.ToSlashPath(DosRelativePath),
+                    StringComparison.OrdinalIgnoreCase);
+        } else {
+            return string.Equals(ConvertUtils.ToSlashFolderPath(relativePath),
+                ConvertUtils.ToSlashFolderPath(DosRelativePath),
+                    StringComparison.OrdinalIgnoreCase);
+        }
```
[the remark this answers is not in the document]
maximilien-noal: Nah.

maximilien-noal:
Ok, this seems to be working now.

But Krondor won't:
- let me restore a game
- create a saves folder and display it in the Save Game UI, like it does with DOSBox.

Until this is fixed, this PR will stay as draft.

src/Spice86.Core/Emulator/OperatingSystem/IDosPathResolver.cs
maximilien-noal: Properties removed.

maximilien-noal:
This fixes #326 and then some.

I'll create another issue for the Restore Game button being unavailable in Krondor.

This PR is Ready for Review.

src/Spice86.Core/Emulator/Memory/IMemory.cs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// <summary>
+    ///     The number of bytes in the memory map.
+    /// </summary>
+    int Size { get; }
+
+    /// <summary>
+    ///     Allows indexed byte access to the memory map.
+    /// </summary>
+    public UInt8Indexer UInt8 { get; }
```
JorisVanEijden, line 125: Is there a difference between `IMemory.UInt8` and `IIndexed.UInt8`?
Perhaps this definition should be removed from one of them?

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -602,119 +543,104 @@ public class DosFileManager {
...
         }
 
-        string fileToProcess = ConvertUtils.ToSlashPath(caseInsensitivePath);
-        string? parentDir = Path.GetDirectoryName(fileToProcess);
-        if (File.Exists(fileToProcess) || Directory.Exists(fileToProcess) ||
-            (string.IsNullOrWhiteSpace(parentDir) == false && Directory.Exists(parentDir) && Directory.GetDirectories(parentDir).Length == 0)) {
-            // file exists or root reached, no need to go further. Path found.
-            return caseInsensitivePath;
+        if (!Directory.Exists(parentFolder)) {
+            CreateDirectory(_dosPathResolver.GetRelativeHostPathToCurrentDirectory(parentFolder));
```
JorisVanEijden, line 576: Why do we need to recurse? Wouldn't `Directory.CreateDirectory(parentFolder)` work here?

maximilien-noal: I don't think it's accepted on all platforms. On Windows, sure. But it may not be so on Unix.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -602,119 +543,104 @@ public class DosFileManager {
...
         }
 
-        string? parent = ToCaseSensitiveFileName(parentDir);
-        if (parent == null) {
-            // End of recursion, root reached
-            return null;
+        if (_dosPathResolver.AnyDosDirectoryOrFileWithTheSameName(dosDirectory, new DirectoryInfo(parentFolder))) {
+            return PathNotFoundError(dosDirectory);
```
JorisVanEijden, line 580: This should be an access_denied error

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,315 @@
...
+    private const char AltDirectorySeparatorChar = '/';
+    private const int MaxPathLength = 255;
+
+    /// <summary>
+    /// Initializes a new instance.
+    /// </summary>
+    /// <param name="configuration">The emulator configuration.</param>
+    public DosPathResolver(Configuration configuration) {
+        _driveMap = InitializeDriveMap(configuration);
+        _currentDrive = 'C';
```
JorisVanEijden, line 26: Maybe move this to the parameters with a default of 'C', or add the default to the configuration?

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
JorisVanEijden: It's a micro-optimization.
If a CDrive is specified in the configuration then we don't need to call `GetExeParentFolder()`.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
JorisVanEijden: That seems rather arbitrary.  I would support it, just in case we later add/allow drive mapping configuration.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
JorisVanEijden: I don't see why not. I would assume the same restrictions apply to `Directory.Exists(parentFolder)`.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -0,0 +1,315 @@
...
+        _driveMap[driveLetter].FullHostCurrentDirectory = hostFullPath;
+        return DosFileOperationResult.NoValue();
+    }
+
+    /// <summary>
+    /// Converts the DOS path to a full host path of the parent directory.<br/>
+    /// </summary>
+    /// <param name="dosPath">The DOS path to convert.</param>
+    /// <returns>A string containing the full path to the parent directory in the host file system, or <c>null</c> if nothing was found.</returns>
+    public string? TryGetFullHostParentPathFromDos(string dosPath) {
```
JorisVanEijden, line 182: In general, `TryGet..` methods return a `bool` to indicate success and return the value via an `out` parameter.
The pattern where a default is returned is usually called `Get...OrDefault`

that's just naming nitpicking though

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    /// <summary>
+    /// The full host path to the mounted folder. This path serves as the root of the DOS drive.
+    /// </summary>
+    public string MountedHostDirectory { get; init; }
+
+    /// <summary>
+    /// The full path to the current host directory in use on the drive.
+    /// </summary>
+    public string FullHostCurrentDirectory { get; set; }
```
JorisVanEijden, line 38: I have a feeling we don't need this one.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    /// <summary>
+    /// The full path to the current DOS directory in use on the drive.
+    /// </summary>
+    public string FullDosCurrentDirectory => $"{DosDriveRootPath}{DosPathResolver.DirectorySeparatorChar}{ConvertUtils.ToBackSlashPath(FullHostCurrentDirectory[MountedHostDirectory.Length..])}".ToUpperInvariant();
```
JorisVanEijden, line 43: Would it not be more logical to only keep track of the current dos directory relative to it's drive letter?

src/Spice86.Core/Emulator/CPU/Stack.cs
```diff
@@ -3,51 +3,51 @@
 using Spice86.Core.Emulator.Memory;
 
 public class Stack {
-    private readonly Memory memory;
+    private readonly IMemory _memory;
```
kevinferrare, line 6: Actually with the PR I just merged, if the class only needs to access UInt8/16/32, it could depend on Indexable instead of IMemory, but I am not sure this is the purpose of your change. Did you change that to mock memory?

maximilien-noal: Exactly.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: I'll rather spend time on it when drive mapping configuration is added, which means adding a real mount system.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: Resolved outdated conversation.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: Fixed. Good catch !

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Method renamed.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: Recursion removed.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/Memory/IMemory.cs
maximilien-noal: I have to move things that are coming from Indexable to a new interface for unit tests, maybe a future IIndexable.

Right now there are type casts of _machine.Memory (which is declared as an instance of IMemory) as an Indexable and I don't like it.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
maximilien-noal: I removed another unused property, and an unused method. But I wasn't able to remove this one.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
maximilien-noal: Moved to a method in DosPathResolver, but it's the same code.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
JorisVanEijden: It should be computed from the mountpoint + the current dos dir. 
Otherwise you're going to always have to synchronize this one and the current dos dir, and you're essentialy stroing the same thing in two different locations.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
maximilien-noal: Done. We only keep track of the current DOS directory, relative to the Drive Letter.

src/Spice86.Core/Emulator/OperatingSystem/MountedFolder.cs
maximilien-noal: This property has been dropped.

src/Spice86.Core/Emulator/Memory/IMemory.cs
maximilien-noal: Removed duplicated UInt8 and such in IMemory.cs

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Done.

PR #359, opened 2023-06-30
Separate Interrupts and Callbacks. Make mouse CALL instruction explicit in memory

kevinferrare:
### Description of Changes
Make a clear distinction between Callbacks and Interrupt handlers:
 - Callbacks are C# code behind a custom ASM instruction
 - Interrupt handlers are C# classes that install themselves in memory as ASM, and that can rely on Callbacks to ensure their C# code is called from emulated ASM

Mouse interrupt handler now works like this:
```Assembly
// Responsible for Regs storage and setup. Does not change stack and CS:IP anymore.
CALLBACK 0x74
// User function address is written by CALLBACK 0x74
CALLF <User Function Address>
// Responsible for regs cleanup as before
CALLBACK 0x90
IRET
// Default user function, does nothing
RETF 
```

Had to change dumped json library because Newtonsoft one did not like SegmentedAddress having 2 constructors and was ignoring [JsonConstructor]

Stopped dynamically replacing callback instruction in RAM. Drawback => Ghidra will complain. Maybe we could export both pristine and tainted RAM images for analysis.


src/Spice86.Core/Emulator/VM/Machine.cs
```diff
@@ -265,47 +268,48 @@ public class Machine : IDisposable {
...
         MouseDriver = new MouseDriver(Cpu, Memory, MouseDevice, machineCreationOptions.Gui, VgaFunctions, machineCreationOptions.LoggerService);
-        var mouseInt33Handler = new MouseInt33Handler(this, machineCreationOptions.LoggerService, MouseDriver);
-        RegisterCallbackHandler(mouseInt33Handler);
-        var mouseIrq12Handler = new BiosMouseInt74Handler(MouseDriver, DualPic, this, machineCreationOptions.LoggerService);
-        RegisterCallbackHandler(mouseIrq12Handler);
-        var mouseCleanupHandler = new CustomMouseInt90Handler(MouseDriver, this, machineCreationOptions.LoggerService);
-        RegisterCallbackHandler(mouseCleanupHandler);
+        Dos = new Dos(this, machineCreationOptions.LoggerService);
 
+        if (Configuration.InitializeDOS is not false) {
```
kevinferrare, line 297: Same behaviour as before, but we need more granularity in configuration here

maximilien-noal: This could be an option group:
https://github.com/commandlineparser/commandline/wiki/Verbs

for example:
Spice86 --initialize bios dos

src/Spice86.Core/Emulator/CPU/VectorTableManager.cs
```diff
@@ -0,0 +1,38 @@
+namespace Spice86.Core.Emulator.CPU;
+
+using Spice86.Core.Emulator.Memory;
+
+/// <summary>
+/// Manages accesses to the vector table
```
maximilien-noal, line 6: Represents the Vector Table

src/Spice86.Core/Emulator/CPU/VectorTableManager.cs
```diff
@@ -0,0 +1,38 @@
...
+/// </summary>
+public class VectorTableManager {
```
maximilien-noal, line 8: VectorTable

src/Spice86.Core/Emulator/Function/Dump/ExecutionFlowDumper.cs
```diff
@@ -3,7 +3,7 @@
 using System.Diagnostics;
 using System.IO;
 
-using Newtonsoft.Json;
+using System.Text.Json;
```
maximilien-noal, line 6: If STJ is *reliable* (ie. doesn't silently fail serialization with an empty JSON object as output, or deserialization with 'null' as ouput), we can keep it.

src/Spice86.Core/Emulator/Function/Dump/RecorderDataWriter.cs
```diff
@@ -60,18 +60,7 @@ public class RecorderDataWriter : RecordedDataIoHandler {
     /// <param name="suffix">The suffix to add to the file name.</param>
     public void DumpMemory(string suffix) {
         string path = GenerateDumpFileName($"MemoryDump{suffix}.bin");
-        File.WriteAllBytes(path, GenerateRelevantRam());
-    }
-
-    /// <summary>
```
maximilien-noal, line 66: Maybe keep it until the new Generator is available ?

We can mark it [Obsolete("reason here")]

src/Spice86.Core/Emulator/InterruptHandlers/Common/IndexBasedDispatcher/IndexBasedDispatcher.cs
```diff
@@ -0,0 +1,86 @@
...
+    public IndexBasedDispatcher(Machine machine, ILoggerService loggerService) {
+        _machine = machine;
+        _loggerService = loggerService;
+    }
+
+    /// <summary>
+    /// Extract the list of runnable items registered
+    /// </summary>
+    /// <returns>A list of all the runnables registered</returns>
+    public IList<T> GetAllRunnables() {
```
maximilien-noal, line 42: Maybe we could keep it as IEnumerable.

LINQ's Deferred execution means that the evaluation of an expression is delayed until its realized value is actually required. It greatly improves performance by avoiding unnecessary execution.

ToList executes the query right now and makes LINQ code less performant. As a general rule, ToList, ToArray and friends should be avoided.

src/Spice86.Core/Emulator/InterruptHandlers/Common/MemoryWriter/MemoryAsmWriter.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <param name="destination">Segmented address of the function to call</param>
+    /// <returns>Returns the address of the call destination segmented address</returns>
+    public SegmentedAddress WriteFarCall(SegmentedAddress destination) {
+        WriteByte(0x9A);
+        return WriteSegmentedAddress(destination);
+    }
+
+    public SegmentedAddress WriteSegmentedAddress(SegmentedAddress address) {
+        // Make a copy of the address since it is going to be modified by our writes.
+        SegmentedAddress ret = new SegmentedAddress(CurrentAddress);
```
maximilien-noal, line 58: If a copy is never meant to be modified, maybe we could have a ValueSegmentedAddress which is a record struct.

(if it can be mutated, then it's simply a struct with the same interface, ie. Segment, Offset, etc...)

Therefore, the copy will live on the stack, and the GC won't have to deal with it.
It will be freed when it goes out of scope.

The emulator creates a lot of SegmentedAddresses.

However, this is probably an over-optimization.

You could also copy the Segment and Offset into a value tuple:

```
(ushort Offset, ushort Segment) = (CurrentAddress.Offset, CurrentAddress.Segment)

File: src/Spice86.Core/Emulator/InterruptHandlers/Common/MemoryWriter/MemoryBytesWriter.cs [hunk 8]
```diff
@@ -0,0 +1,24 @@
+namespace Spice86.Core.Emulator.InterruptHandlers.Common.MemoryWriter;
+
+using Spice86.Core.Emulator.Memory;
+using Spice86.Shared.Emulator.Memory;
+
+public class MemoryBytesWriter {
```
maximilien-noal, line 6: Missing XML documentation.

src/Spice86.Core/Emulator/InterruptHandlers/InterruptInstaller.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using Spice86.Core.Emulator.CPU;
+using Spice86.Core.Emulator.Function;
+using Spice86.Core.Emulator.InterruptHandlers.Common;
+using Spice86.Core.Emulator.InterruptHandlers.Common.Callback;
+using Spice86.Core.Emulator.InterruptHandlers.Common.MemoryWriter;
+using Spice86.Core.Emulator.Memory;
+using Spice86.Shared.Emulator.Memory;
+
+public class InterruptInstaller {
```
maximilien-noal, line 11: Missing XML documentation

src/Spice86.Core/Emulator/ReverseEngineer/CSharpOverrideHelper.cs
maximilien-noal, line 1: Will have to test it with projects defining overrides (ie. Cryogenic)

Is there a migration path ? Is a migration path needed at all, or is it source compatible ?

maximilien-noal (commented):
This clearly is a much needed improvement.
It really clears things up ! I love it.
I left a few comments / puzzling questions.

src/Spice86.Core/Emulator/VM/Machine.cs
kevinferrare: Created an issue to keep track of this, to do in a separate PR: https://github.com/OpenRakis/Spice86/issues/361

src/Spice86.Core/Emulator/Function/Dump/ExecutionFlowDumper.cs
kevinferrare: Would have kept Newtonsoft but I couldnt make it work with SegmentedAddress having 2 constructors even after specifying [JsonConstructor]

src/Spice86.Core/Emulator/CPU/VectorTableManager.cs
kevinferrare: It's not the actual table, just a wrapper. Will rework the class to make that clearer

src/Spice86.Core/Emulator/InterruptHandlers/Common/IndexBasedDispatcher/IndexBasedDispatcher.cs
kevinferrare: Good idea, I am a bit lost with LINQ, need more practice.

src/Spice86.Core/Emulator/InterruptHandlers/Common/MemoryWriter/MemoryAsmWriter.cs
kevinferrare: Would be nice for general use, we have a lot of places where we deal with segement / offset params separately to avoid an allocation. However here this code is called only a few times before execution starts so it's not going to break GC :D

src/Spice86.Core/Emulator/InterruptHandlers/Common/MemoryWriter/MemoryBytesWriter.cs
kevinferrare: Done

src/Spice86.Core/Emulator/InterruptHandlers/InterruptInstaller.cs
kevinferrare: Done

src/Spice86.Core/Emulator/ReverseEngineer/CSharpOverrideHelper.cs
kevinferrare: Good catch!
This is for sure not API compatible at least because SetProvidedInterruptHandlersAsOverridden got removed.
This function was broken and could not work for mouse anyway.

There are 2 cases:
 - When no generation is involved, we won't be able to call InterruptCall(...,...,0x21) to call emulator provided interrupts in our overrides. But we don't do that because hand written code can call Dos.Blabla instead :)
 - For generator v1, it won't work well because CALLBACK is not understood by ghidra. User would have to provide code for handlers manually. I could either:
  - Reimplement a smarter SetProvidedInterruptHandlersAsOverridden (but that would be a bit complex because it would need to behave differently for mouse, ems and regular int)
  - Make the generator issue a call to Callback when encountering and INT call in segments > F000 (hacky!)
  - Do nothing :)

src/Spice86.Core/Emulator/CPU/VectorTableManager.cs
```diff
@@ -0,0 +1,38 @@
...
+        _memory.UInt16[offsetAddress] = offset;
+        _memory.UInt16[offsetAddress + 2] = segment;
+    }
+
+    /// <summary>
+    /// Reads the vector table at the given index (vectorNumber) and returns the address of the handler
+    /// </summary>
+    /// <param name="vectorNumber">vector to read</param>
+    /// <returns>A pair of ushort values that are segment and offset</returns>
+    public (ushort segment, ushort offset) ReadVectorFromTable(byte vectorNumber) {
```
JorisVanEijden, line 34: This functionality is also duplicated in `Spice86.Core.Emulator.InterruptHandlers.VGA.VgaFunctionality.GetInterruptVectorAddress` and `Spice86.Core.Emulator.ReverseEngineer.CSharpOverrideHelper.InterruptCall`

kevinferrare: Nice catch! This means this class was even more needed.
I refactored it to use an indexer and tuples.

src/Spice86.Core/Emulator/CPU/VectorTableManager.cs
kevinferrare: Rewrote the class with an indexer, much cleaner.

maximilien-noal (approved):

PR #362, opened 2023-07-02
Breakpoints cleanup

kevinferrare:
Remove range breakpoints that were insanely slow
Tried to replace the dictionary in the BreakpointHolder class with an array, but it didnt lead to any concrete performance increase.


maximilien-noal (approved):
Nice cleanup !

src/Spice86.Core/Emulator/Memory/Memory.cs
```diff
@@ -63,15 +63,12 @@ public class Memory {
     }
 
     /// <summary>
-    ///     Writes a 4-byte value to ram.
+    ///     Writes a 1-byte value to ram.
```
kevinferrare, line 66: Reorganized the methods order because it bothered me a lot :)

maximilien-noal (approved):

PR #366, opened 2023-07-05
Migration/avalonia v11

maximilien-noal:
### Description of Changes

Migration to the latest version of the UI framework, which was published a few hours ago.

### Rationale behind Changes

Eventually, the v0.10.x version will not be supported anymore.

Plus, the performance regressions found in AvaloniaUI v11-rc1 were fixed.

### Suggested Testing Steps

Test this while looking at the average number of instructions per second, compared with the master branch.

I already did so on too very different machines, and the numbers were very close.

### DIFF
--- a/src/Spice86/App.axaml
+++ b/src/Spice86/App.axaml
@@ -4,9 +4,6 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:Spice86">
     <Application.Styles>
-        <FluentTheme Mode="Light" />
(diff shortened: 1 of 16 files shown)


maximilien-noal:
Some UI Commands and the keyboard don't work. Back to draft.

maximilien-noal:
Keyboard issue fixed, now only some RelayCommands have non-working XAML bindings.

maximilien-noal:
I had to remove the CommandParameter usage in order to fix them.

kevinferrare (approved):

PR #369, opened 2023-07-07
Rework Memory class

kevinferrare:
Remove Get / Set methods, use indexers everywhere.
Make indexers depend on IByteReaderWriter instead of Memory.
Create a new IIndexed interface to represent Indexer 8 / 16 / 32 holders.

Motivation: to be able to use a byte array as IIndexed for memory dumps.


maximilien-noal (commented):
Every line is fine. This streamlines the memory accesses a lot !

maximilien-noal (approved):

maximilien-noal (approved):

PR #382, opened 2023-07-13
Fix/perf UI and startup code

### LINKED ISSUE #328: Performance investigations
Try and make the BreakPointHolder use a List spanning the entire Memory instead of a Dict

maximilien-noal:
### Description of Changes

Makes the startup code faster. We only parse the command line once, for example.

Also, the performance window calculations were wrong. This is fixed.

### Rationale behind Changes

The startup code was overly complicated. So was the performance window view model. 

### Suggested Testing Steps

Startup both the UI mode and the headless mode and see if they work.


kevinferrare (approved):
Pause system needs refactor, but that's a story for another day.
Not sure to be able to comment on UI parts.

PR #385, opened 2023-07-21
Uncouple INT74 and mouse driver implementations. Mouse driver now installs it's assembly code in ram itself and handles user routine call. INT74 calls the mouse driver when present.

kevinferrare:
Uncouple INT74 and mouse driver implementations.
Mouse driver now installs it's assembly code in ram itself and handles user routine call.
INT74 calls the mouse driver when present.


src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseDriver.cs
```diff
@@ -251,4 +258,27 @@ public class MouseDriver : IMouseDriver {
...
+        //  - Create a callback (0xFE) that will call the BeforeUserHandlerExecution method
+        //  - Create a Far call instruction that is calling a dummy handler by default, but can be modified to call anything
+        //  - Create a callback (0xFF) that does the cleanup
+        //  - Create a FAR ret
+        //  - Make the far call point to the FAR ret so that without any setup we call a user routine that does nothing
+        
+        // Write ASM
+        SegmentedAddress driverAddress = new SegmentedAddress(memoryAsmWriter.CurrentAddress);
+        memoryAsmWriter.RegisterAndWriteCallback(BeforeUserHandlerExecutionCallbackNumber, BeforeUserHandlerExecution);
+        // Far call to nothing for now, we will overwrite the address to the default handler later
```
maximilien-noal, line 274: 'Does nothing for now' ?

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseDriver.cs
```diff
@@ -251,4 +258,27 @@ public class MouseDriver : IMouseDriver {
     }
 
     private record MouseRegisters(ushort Es, ushort Ds, ushort Di, ushort Si, ushort Bp, ushort Sp, ushort Bx, ushort Dx, ushort Cx, ushort Ax);
```
maximilien-noal, line 260: If not mutable, use a 'readonly record'

maximilien-noal (approved):

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseDriver.cs
kevinferrare: Reworked the code to make the intent less cryptic...

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseDriver.cs
```diff
@@ -88,19 +88,28 @@ public class MouseDriver : IMouseDriver {
         return true;
     }
 
-    private void CallUserRoutine() {
+    private void PrepareUserRoutineCall() {
         // Save registers so that we can restore them later when user routine is done
         SaveRegisters();
         // User handler is going to be called. Save registers, set them to their new values
         SetRegistersToMouseState();
         // Set the address of assembly routine to call. Set it every time to ensure it is what we want
```
JorisVanEijden, line 96: Leftover commernt

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseDriver.cs
```diff
@@ -63,16 +66,13 @@ public class MouseDriver : IMouseDriver {
     /// <inheritdoc />
     public int CurrentMinY { get; set; }
 
-    /// <inheritdoc />
-    public IAsmUserRoutineHandler? UserRoutineHandler { set; private get; }
-
     /// <inheritdoc />
     public void BeforeUserHandlerExecution() {
```
JorisVanEijden, line 70:
```suggestion
    public void HandlePS2MouseDataUpdate() {
```
The purpose of the call to this method is to let the mouse driver know that the PS/2 mouse device has new data.
We currently only have 1 mouse device (PS/2) and that is cheating by not putting the data on the I/O ports but exposing it through public properties.
If we want more accurate emulation and more device types we are going to need to handle input from more sources, triggered by other things than irq12/int74.
Only after processing this data, and making it available to int33 queries, the driver should check if someone else has expressed an interest in mouse events by registering a usercallback. Only then, if the registration matches the current event, should a call be made to the user code.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseDriver.cs
```diff
@@ -251,4 +260,28 @@ public class MouseDriver : IMouseDriver {
     }
 
     private record MouseRegisters(ushort Es, ushort Ds, ushort Di, ushort Si, ushort Bp, ushort Sp, ushort Bx, ushort Dx, ushort Cx, ushort Ax);
+    
+    /// <inheritdoc />
+    public SegmentedAddress WriteAssemblyInRam(MemoryAsmWriter memoryAsmWriter) {
```
JorisVanEijden, line 265: I'm not a big fan of setting all this up in advance, just in case a game decides to register a mouse event handler.
Ideally this is done on `RegisterCallback()`.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseDriver.cs
kevinferrare: Indeed, done

JorisVanEijden (approved):

PR #414, opened 2023-08-10
Fix SoundBlaster not consuming data when audio backend cannot be created. Display an error in this case and continue without sound.

kevinferrare:


maximilien-noal (changes requested):
The backend. Audio.AudioPlayer should not return null and the callers should not check for null.

Instead, it should return a NullAudioPlayer class that does nothing.

This way this fixes the problem for all present (OPL3FM, general MIDI, MT-32, PC speaker) and future (Adlib Gold, Gravis Ultrasound, Covox Speech Thing, Disney Sound Source) sound devices.

maximilien-noal (approved):

PR #415, opened 2023-08-12
Make code easier to use for override suppliers

kevinferrare:


src/Spice86.Core/Emulator/Memory/Indexer/Indexer.cs
```diff
@@ -22,6 +22,15 @@ public abstract class Indexer<T> {
         get => this[(uint)address];
         set => this[(uint)address] = value;
     }
+    
+    /// <summary>
+    /// Indexer for addresses that are long. For convenience.
+    /// </summary>
+    /// <param name="address"></param>
+    public T this[long address] {
```
kevinferrare, line 30: this is because uint + int is long and sometimes address calculations end up like this

JorisVanEijden (approved):

maximilien-noal (approved):

PR #438, opened 2023-09-13
Feature/quick debug

maximilien-noal:
### Description of Changes

In progress implementation of an internal "debugger" of sorts.

This is just a quick tool, won't be big and professional like GDB.

### Rationale behind Changes

Debuggability is improved. The UI code is also unit test friendly, with interfaces everywhere it made sense, and some refactorings.

### Suggested Testing Steps

Test out the UI while the emulator is paused. I already did, of course.

### Future plans

Required:

- Display the Call stack
- Display the return address
- Display the dissasembly (with the help of ICED)

Maybe:

- Memory "Go to location..." function, Memory editing, Memory byte sequence search (with hexadecimal values as input)
- Debug control (pause/step/continue)

Bonus tier:

- Display content of mapped and unmapped EMS pages


src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
```diff
@@ -35,7 +37,9 @@ public class Timer : DefaultIOPortHandler {
     // screen refresh
     private readonly Counter _vgaScreenRefreshCounter;
 
-    public Timer(State state, ILoggerService loggerService, DualPic dualPic, IVideoCard? vgaCard, CounterConfigurator counterConfigurator, bool failOnUnhandledPort) : base(state, failOnUnhandledPort, loggerService) {
+    private readonly IGui? _gui;
+
+    public Timer(IGui? gui, State state, ILoggerService loggerService, DualPic dualPic, IVideoCard? vgaCard, CounterConfigurator counterConfigurator, bool failOnUnhandledPort) : base(state, failOnUnhandledPort, loggerService) {
```
JorisVanEijden, line 42: This does not feel right, the Timer depending on the Gui, rather than the Gui depending on the Timer.
Is there a specific reason you set it up this way?

src/Spice86/Views/DebugWindow.axaml
```diff
@@ -18,455 +20,468 @@
...
-            </Grid>
-            <TextBlock Text="To show anything, pause the emulator first." IsVisible="{Binding !IsPaused}" />
-        </Grid>
-    </ScrollViewer>
+                <Button Margin="5,0,0,0" VerticalAlignment="Bottom" Content="Force Update"
+                        Command="{Binding UpdateDataCommand}">
+                </Button>
+            </WrapPanel>
+        </Border>
+        <TextBlock Grid.Row="0" Grid.RowSpan="2" Text="To show anything, pause the emulator first." IsVisible="{Binding !IsPaused}" />
```
JorisVanEijden, line 485: I'd really like to keep seeing the video registers change while the game is running. Having to pause it is a dealbreaker for me.

JorisVanEijden (changes requested):
Can you make the debug widnows read-only while running and editable while paused?
That would be awesome.

src/Spice86/Views/DebugWindow.axaml
maximilien-noal: I'll fix this. This was to avoid reads while the emulator changes state (once I had an exception while showing the values of the DAC. In other words, reading the ArgbPalette while its length changed mid-read lead to an unhandled IndexOutOfRangeException).

I agree that seeing values withotu having to pause is valuable. I think I'll just catch and ignore IndexOutOfRangeException in the DebugViewModel code that updates the data presented.

src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
maximilien-noal: Kevin said the same thing. This was to avoid any Law of Demeter violations in the ProgramExecutor. Most where resolved by the Visitor pattern, but not this one.

I didn't like the idea of using the Visitor pattern for everything. I'll think of another way (maybe a nullable ITimeMultiplier property that the Timer sets on creation).

The real fix would be to inject the timer into the GUI, but the the Timer depends on so many services that it becomes nightmarish to try and move all instanciations into Program.cs (where we create MainWindowViewModel).

maximilien-noal:
> Can you make the debug widnows read-only while running and editable while paused? That would be awesome.

The PropertyGrid supports that, and the rest could also support it. I thought about it, but I didn't know how valuable it was. I think I can support it easily with the Visitor pattern. It will not only read but also mutate state. Quite dangerous for maintainability, I think. A good thing it's only used in one place. :)

src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
maximilien-noal: I now recall that I saw a WPF ChatGPT example where Microsoft's DI (IServicesCollection) was used with the WPF app host. I can do something similar with the AvaloniaUI app host and inject the Timer into the UI.

This could even potientally remove the Visitor pattern... Must think more about this. :)

src/Spice86/Views/DebugWindow.axaml
maximilien-noal: Done.

src/Spice86.Core/Emulator/Devices/Timer/Timer.cs
maximilien-noal: Fixed by using a property in IGui than the Machine sets once the Timer is created. The IGui doesn't inherit from INotifyPropertyChanged anymore either.

maximilien-noal:
> Can you make the debug widnows read-only while running and editable while paused?

This is done (for the CPU tab, at least. It could be expanded)

JorisVanEijden:
A couple of things:
- when the debug window is open and I Pause and Continue one or more times, an Invalid Opcode exception occurs. I think there may be some race conditions between the GUI's state and the actual state.
- my monitor is not large enough to fit the Video debug info anymore :) Honestly I don't think there's a usecase for altering the VGA registers. And it's a nigthmare to sync, since the VGA continues to fire, even in Paused state.
- memory editing would be the most useful, but not required for this PR
- if the memory grid could show segment:offset addresses that would be super useful

maximilien-noal:
Oh crap I thought I had fixed that by unregistering the OnPropertyChanged event handler when not Paused. I'll fix it right now, along with the rest.

Thank you for your feedback !

JorisVanEijden (approved):

src/Spice86.Core/Emulator/VM/Machine.cs
```diff
@@ -346,4 +349,12 @@ public sealed class Machine : IDisposable {
         Dispose(disposing: true);
         GC.SuppressFinalize(this);
     }
+
+    public void Accept(IEmulatorDebugger emulatorDebugger) {
+        Cpu.Accept(emulatorDebugger);
+        ((IDebuggableComponent)VgaCard).Accept(emulatorDebugger);
```
kevinferrare, line 355: Is cast needed? It seems VgaCard implements IDebuggableComponent isnt it?

src/Spice86.Core/Emulator/Memory/Memory.cs
```diff
@@ -26,13 +25,18 @@ public class Memory : Indexable.Indexable, IMemory {
...
+    public Memory(IMemoryDevice baseMemory, bool is20ThAddressLineSilenced, bool initializeResetVector) {
         uint memorySize = baseMemory.Size;
         _memoryDevices = new IMemoryDevice[memorySize];
         Ram = new Ram(memorySize);
         RegisterMapping(0, memorySize, Ram);
         (UInt8, UInt16, UInt32, Int8, Int16, Int32, SegmentedAddressValue, SegmentedAddress) = InstantiateIndexersFromByteReaderWriter(this);
         A20Gate = new(is20ThAddressLineSilenced);
+        if(initializeResetVector) {
+            // Put HLT at the reset address
+            UInt16[0xF000, 0xFFF0] = 0xF4;
```
kevinferrare, line 38: How about putting that in Machine since it does checks on whether it's DOS or BIOS?

src/Spice86.Core/Emulator/VM/Machine.cs
maximilien-noal: I don't like it either (too many parentheses, makes the code harder to read), however the types are declared as IVideoCard, IMemory and such, that's why the cast is needed.

JetBrains Rider doesn't detect unneeded casts.

src/Spice86.Core/Emulator/Memory/Memory.cs
maximilien-noal: Will do.

src/Spice86.Core/Emulator/VM/Machine.cs
maximilien-noal: Maybe the IVideoCard, IMemory etc... interfaces could inherit from the IDebuggableComponent interface ?

src/Spice86/Models/Debugging/CpuStateInfo.cs
```diff
@@ -0,0 +1,84 @@
...
+    [ObservableProperty] private ushort _CS;
+    [ObservableProperty] private ushort _CX;
+
+    [ObservableProperty, ReadOnly(true)] private long _cycles;
+
+    [ObservableProperty] private byte _DH;
+
+    [ObservableProperty] private ushort _DI;
+
+    [ObservableProperty] private short _direction16;
```
kevinferrare, line 34: Not sure direction is needed in the UI, it is a derived value from _directionFlag

maximilien-noal: Editing them directly does not make sense ?

kevinferrare: No, and in fact they are calculated properties:
```
public short Direction8 => (short)(DirectionFlag ? -1 : 1);
```

src/Spice86.Core/Emulator/VM/Machine.cs
kevinferrare: Yes that would be the best solution

src/Spice86/Models/Debugging/CpuStateInfo.cs
maximilien-noal: Fixed.

src/Spice86.Core/Emulator/Memory/Memory.cs
maximilien-noal: Fixed on master.

PR #441, opened 2023-09-22
Alu class split between 8, 16 and 32 versions

kevinferrare:
Split the ALU between 8, 16 and 32bits versions with a common class Alu providing common logic and abstract methods.
This is needed to unify instructions behaviour with generic math in cfgcpu.


maximilien-noal (approved):
Very good improvement, everything is clear cut, approved !

JorisVanEijden (approved):

maximilien-noal:
Dune, Hero Quest, and Krondor tested without any problem.

PR #442, opened 2023-09-24
Use indexers to access registers

kevinferrare:


src/Spice86.Core/Emulator/CPU/ModRM.cs
```diff
@@ -18,48 +19,48 @@ public class ModRM {
         _state = state;
     }
 
-    public uint GetAddress(int defaultSegmentRegisterIndex, ushort offset) {
-        int? segmentIndex = _state.SegmentOverrideIndex;
+    public uint GetAddress(uint defaultSegmentRegisterIndex, ushort offset) {
+        uint? segmentIndex = _state.SegmentOverrideIndex;
         segmentIndex ??= defaultSegmentRegisterIndex;
```
JorisVanEijden, line 24:
```suggestion
        uint? segmentIndex = _state.SegmentOverrideIndex ?? defaultSegmentRegisterIndex;
```

src/Spice86.Core/Emulator/CPU/ModRM.cs
```diff
@@ -18,48 +19,48 @@ public class ModRM {
...
 
-        ushort segment = _state.SegmentRegisters.GetRegister16((int)segmentIndex);
+        ushort segment = _state.SegmentRegisters.UInt16[(int)segmentIndex];
```
JorisVanEijden, line 26: Why is a `uint?` cast to an `int`?
I'd expect the segmentIndex to either already be an `int`, or the `segmentIndex.Value` being used.

src/Spice86.Core/Emulator/CPU/Registers/UInt32RegistersIndexer.cs
```diff
@@ -0,0 +1,19 @@
...
+public class UInt32RegistersIndexer : RegistersIndexer<uint> {
+    private readonly IUIntReaderWriter _uIntArrayReaderWriter;
+
+    public UInt32RegistersIndexer(IUIntReaderWriter uIntArrayReaderWriter) {
+        _uIntArrayReaderWriter = uIntArrayReaderWriter;
+    }
+
+    public override uint this[uint index] {
+        get => _uIntArrayReaderWriter[(uint)index];
+        set => _uIntArrayReaderWriter[(uint)index] = value;
```
JorisVanEijden, line 17: Isn't index already an uint?

kevinferrare: indeed, there are lot of places like that in the code, Indexer actually accepts both uint and int

src/Spice86.Core/Emulator/CPU/ModRM.cs
kevinferrare: Done

JorisVanEijden (approved):

maximilien-noal (approved):

PR #495, opened 2023-11-21
ExecutionFlowRecorder: Fixed mini GC Storm

maximilien-noal:
### Description of Changes

I know this is going to be replaced by the CFG CPU, but:

We cache the action in a static field, this allows the onReached action to be a static lambda. This avoid a lot of Actions being created on the GC Heap to be reclamed by the GC later.

We also don't pass the AddressBreakPoint as a parameter, as it is being created. The only way to pass it would be to make the action non static, and we don't want that.

We discard the paramater ("_" is the discard action in C#) of the Invoke method, since we *are* creating the breakpoint.

To allow this change, BreakPoint.Action has become mutable (get; set; instead of get; private set;).

### Rationale behind Changes

GC pressure or GC storms is what kills performance in an application.

Before:

image

After:

image

### Suggested Testing Steps

Test that the dump created on disk by the ExecutionFlowRecorder in the same before and after the change. It should be.

### DIFF
--- a/src/Spice86.Core/Emulator/Function/ExecutionFlowRecorder.cs
+++ b/src/Spice86.Core/Emulator/Function/ExecutionFlowRecorder.cs
@@ -96,7 +96,7 @@ public class ExecutionFlowRecorder {
         // Beyond that, this code would have to be deleted,
         // as the allocation would take way too much time.
         for (uint i = 0; i < Memory.A20Gate.EndOfHighMemoryArea; i++) {
-            _addressBreakPoints.Add(i, GenerateBreakPoint(memory, state, i));
+            _addressBreakPoints.Add(i, GenerateBreakPoint(this, memory, state, i));
         }
     }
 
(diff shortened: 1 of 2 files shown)


JorisVanEijden (changes requested):
Uhm, this looks strange. This seems to create the situation wherein the 1st generated breakpoint is executed, no matter which breakpoint is hit.

I also do not understand the graphs. The 1st shows a lot of live objects but hardly any delta. The 2nd show a lot less live objects but much more activity in creating and destroying objects.

maximilien-noal:
Yes I tough about it overnight and this does not fly.

Since this will be replaced by the cfg cpu, I will either find a quick fix or abandon this PR.

maximilien-noal:
> Uhm, this looks strange. This seems to create the situation wherein the 1st generated breakpoint is executed, no matter which breakpoint is hit.
> 
> I also do not understand the graphs. The 1st shows a lot of live objects but hardly any delta. The 2nd show a lot less live objects but much more activity in creating and destroying objects.

I fixed his with a Dictionary<AddressBreakPoint, Action<BreakPoint>>

kevinferrare (dismissed):

src/Spice86.Core/Emulator/Function/ExecutionFlowRecorder.cs
```diff
@@ -208,22 +208,27 @@ public class ExecutionFlowRecorder {
...
 
-            byte oldValue = memory.UInt8[physicalAddress];
-            byte newValue = memory.CurrentlyWritingByte;
-            if (oldValue != newValue) {
-                RegisterExecutableByteModification(
-                    new SegmentedAddress(state.CS, state.IP), physicalAddress, oldValue, newValue);
-            }
-        }, false);
+    private AddressBreakPoint GenerateBreakPoint(IMemory memory, State state, uint physicalAddress) {
+        _onReached ??= _ => OnBreakPointReached(this, memory, state, physicalAddress);
```
kevinferrare, line 214: Won't this erase addresses of previously generated breakpoints?

maximilien-noal:
image
Not worth it, PR abandonned.

PR #528, opened 2023-12-23
More Performance

maximilien-noal:
### Description of Changes

Makes the ExecutionFlowRecorder (and therefore the CPU) way faster.

To do this, It introduces a breaking change: The ExecutedInstructions HashSet (which used the _instructionsEncountered HashSet as a Key column) is now a Dictionary. The change is a breaking one because deserialization of previously JSON-serialized ExecutionFlowRecorder instances won't work anymore. They will have to be discarded.

Also contains some micro-optimizations that felt good, but didn't change anything (like a static lambda for the UI video refresh, or the introduction of more FrozenSet/FrozenDictionary for lookup tables) performance wise.

Measures:

8M to 9M instructions per second average at most on master, measured at 84M executed instructions
11M to 12M instructions per second average very quickly with this fix, measured at 4M executed instructions.

image


maximilien-noal:
Measurments of last night are not reproduced this morning, converted to draft.

PR #558, opened 2024-01-18
More/performance

maximilien-noal:
### Description of Changes

Performance focused changes all over the code base: GUI, ExecutionFlowRecorder, Mouse Driver, sound, and some other places.

### Rationale behind Changes

Small performance related changes which, will they don't change the perf numbers, they give further ideas for future investigation and do remove some very minor GC pressure (with static lambdas, for example).

### Suggested Testing Steps

Test this with your favorite game. I already did, no regressions found.


src/Spice86.Core/Emulator/VM/EmulationLoop.cs
```diff
@@ -117,28 +137,34 @@ public class EmulationLoop {
...
-        if (_gdbCommandHandler is null) {
+        if(_gdbCommandHandler == null) {
             return false;
         }
-
         _gdbCommandHandler.Step();
         return true;
     }
-    
-    private void PauseIfAskedTo() {
```
kevinferrare, line 133: I would keep the method so that the if in RunLoop makes sense here

maximilien-noal: Done.

kevinferrare (approved):

Issue #212, opened 2022-12-04
Where In Space Is Carmen Sandiego black screen


Die4Ever:
**Describe the bug**
for Where In Space Is Carmen Sandiego, I only see a black screen, but it shows instructions are running

**To Reproduce**
You can easily reproduce this with the shareware demo 

https://archive.org/details/msdos_WISPACE_shareware - direct download link: https://archive.org/download/msdos_WISPACE_shareware/WISPACE.zip

```
C:\Users\die4e\Downloads\Spice86 Release\Release\net7.0\win-x64>Spice86.exe -e "C:\Games\carmen sandiego\msdos_WISPACE_shareware\WISPACE\demo.exe"
[2022-12-03 23:10:58.285 -06:00 ERR {"SourceContext": "Spice86.Core.Emulator.InterruptHandlers.Dos.DosMemoryManager"}] Could not find any MCB to fit 65535.
[2022-12-03 23:10:58.314 -06:00 ERR {"SourceContext": "Spice86.Core.Emulator.InterruptHandlers.Dos.DosInt21Handler"}] DOS operation failed with an error. Int will return to 0x143B:0x2A76/0x16E26. . State is Cycles=2668 CS:IP=0xF000:0x20/0xF0020 EAX=0x48A4 EBX=0xFFFF ECX=0x200 EDX=0xFFFF ESI=0x32A8 EDI=0x32A8 EBP=0x81DC ESP=0x80CE SS=0x2972 DS=0x2972 ES=0xFFFF FS=0x0 GS=0x0 flags=0x7086 (    S  P )
[2022-12-03 23:10:58.327 -06:00 ERR {"SourceContext": "Spice86.Core.Emulator.InterruptHandlers.Dos.DosMemoryManager"}] MCB {"FileName": "", "PspSegment": 4080, "Size": 28266, "TypeField": 90, "UsableSpaceSegment": 12693, "IsFree": false, "IsLast": true, "IsNonLast": false, "IsValid": true, "BaseAddress": 203072, "Memory": {"CurrentlyWritingByte": 114, "Ram": "37013223000000000000000000000000... (1048576 bytes)", "Size": 1048576, "UInt8": {"$type": "UInt8Indexer"}, "UInt16": {"$type": "UInt16Indexer"}, "UInt32": {"$type": "UInt32Indexer"}, "$type": "Memory"}, "$type": "DosMemoryControlBlock"} is too small for requested size 65535.
[...]

PR #562, opened 2024-01-22
Add Bios INT 12h to report installed memory size.

JorisVanEijden:
### Description of Changes
Add Bios INT 12h to report installed memory size.

### Rationale behind Changes
https://github.com/OpenRakis/Spice86/issues/212 mentioned a game using it and we didn't have it yet.

### DIFF
--- a/src/Spice86.Core/Emulator/VM/Machine.cs
+++ b/src/Spice86.Core/Emulator/VM/Machine.cs
@@ -130,6 +130,11 @@ public sealed class Machine : IDisposable, IDebuggableComponent {
     public SoundBlaster SoundBlaster { get; }
 
     /// <summary>
+    /// INT12H handler.
+    /// </summary>
+    public SystemBiosInt12Handler SystemBiosInt12Handler { get; }
+
+    /// <summary>
     /// INT15H handler.
     /// </summary>
     public SystemBiosInt15Handler SystemBiosInt15Handler { get; }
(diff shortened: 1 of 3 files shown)


maximilien-noal:
This is exactly the kind of code that has been piling up in fix/detroit. I approve !

Lots of little DOS / BIOS interrupt handlers missing, and this is one of them.

(Dune is a very relaxed game on that front :D )

maximilien-noal (approved):

src/Spice86.Core/Emulator/VM/Machine.cs
```diff
@@ -200,7 +205,9 @@ public sealed class Machine : IDisposable, IDebuggableComponent {
             Memory.UInt16[0xF000, 0xFFF0] = 0xF4;
         }
         IoPortDispatcher = ioPortDispatcher;
-        BiosDataArea = new BiosDataArea(Memory);
+        BiosDataArea = new BiosDataArea(Memory) {
+            MemSizeKb = (ushort)Math.Clamp(Memory.Ram.Size / 1024, 0, 1024) // max 1mb.
```
kevinferrare, line 209: Looks like this is not supposed to include ram above 640k
http://www.techhelpmanual.com/184-int_12h__conventional_memory_size.html
0x280

JorisVanEijden: Thanks, fixed. 
My source only said to use other interrupts for memory above 1mb. 640K makes more sense.

PR #632, opened 2024-04-06
Fix/audio stack performance

maximilien-noal:
Fixes terrible audio performance on low end machines, such as a Thinkpad T460 or horrible VMs.

Tested with Dune.


maximilien-noal:
Voices are weird, probably a fault on my end in the SoundBlaster class when reverting the code.

kevinferrare (approved):

PR #694, opened 2024-05-31
fix: oversight in SoftwareMixer loop

maximilien-noal:


src/Spice86.Core/Emulator/Devices/Sound/SoftwareMixer.cs
```diff
@@ -42,7 +42,7 @@ public sealed class SoftwareMixer : IDisposable, IDebuggableComponent {
         float finalVolumeFactor = volumeFactor * (1 + separation);
         
         Span<float> target = stackalloc float[data.Length];
-        for (int i = 0; i < data.Length; i++) {
+        for (int i = 0; i < data.Length; i += 2) {
```
JorisVanEijden, line 45: Are you sure you only want to process the _even_ entries in `data`?

maximilien-noal: Fixed, the oversight was bigger than I understood on first read.

This was a leftover from when the code was using the imported AudioFrame struct from DOSBox Staging.

JorisVanEijden (approved):
Not an expert on audio, but this seems simpler than it was.

PR #739, opened 2024-06-20
fix/portaudio_github_actions

maximilien-noal:
by testing if the GITHUB_ACTIONS env var is defined.


maximilien-noal:
Closed issue, as the workflow seems to work reliably now since the last fix.

PR #740, opened 2024-06-20
fix: log template didn't include exception details coming from Serilog.Exception Enricher (apart from base properties such as the StackTrace, put by Serilog itself)

maximilien-noal:


src/Spice86.Logging/LoggerService.cs
```diff
@@ -11,7 +11,7 @@ public class LoggerService : ILoggerService {
     /// <summary>
     /// The format for the log message that will be output.
     /// </summary>
-    private const string LogFormat = "[{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz}] [{Level:u4}] [{IP:j}] {Message:lj}{NewLine}{Exception}";
+    private const string LogFormat = "[{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz}] [{Level:u4}] [{IP:j}] {Message:lj}{NewLine}{Exception} {Properties:j}";
```
JorisVanEijden, line 14: This is adding an empty `{}` to the start of every logline.
Makes sens since your jsonfying an empty object after the NewLine.
But that's probably not what we want.

maximilien-noal: Probably because none of our exceptions add properties to the base class ?

Anyway, this change is described as required by the serilog exception enricher's readme.

JorisVanEijden: It makes the logging like this:
```
[2024-06-28 19:36:43.051 +02:00] [WARN] [36BC:028C] Some logging
 {}[2024-06-28 19:36:43.054 +02:00] [WARN] [36BC:028C] Some more logging
 {}[2024-06-28 19:36:43.057 +02:00] [EROR] [36BC:028C] Even more logging 
 {}
```
Notice the ` {}` in front of every line? That's the ` {Properties:j}`.

And when we log an exception it looks like
```
[2024-06-28 19:36:40.986 +02:00] [EROR] [0000:0000] Example error message
System.Exception: example
 {"ExceptionDetail": {"Type": "System.Exception", "HResult": -2146233088, "Message": "example", "Source": null}}[2024-06-28 19:36:42.822 +02:00] [INFO] [0000:0000] Mouse initialized: Ps2
```
You get the extra properties, but the next log message is appended directly behind it so it gets kinda lost.

The "required change" from that repo just isn't very good.

Do we even have a use-case for logging extended exception properties? Otherwise we might as well get rid of that exception enricher package.

If we do, then we may want to look into using [Serilog.Expressions](https://github.com/serilog/serilog-expressions) to get more flexibility in the log output. See also https://stackoverflow.com/questions/56902157/in-serilog-how-can-i-remove-empty-brackets-from-json-formatted-log-messages-whe

PR #768, opened 2024-07-06
fix: test if we can load PortAudio before trying

maximilien-noal:
Hopefully FINALLY fixes the github windows workflow!


maximilien-noal:
The crash is unrelated to portaudio. Closing this.

PR #792, opened 2024-07-28
Pause handling

JorisVanEijden:
### Description of Changes
Centralize and inject the PauseHandler so all threads, except Main and UI, can be paused.

### Rationale behind Changes
I wanted to trigger a pause from my override code and then be able to resume it from the UI.

### Suggested Testing Steps
Play games, pause, debug, breakpoint, etc.

### DIFF
--- a/src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
+++ b/src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
@@ -108,6 +108,15 @@ public class DosInt21Handler : InterruptHandler {
         AddAction(0x4F, () => FindNextMatchingFile(true));
         AddAction(0x51, GetPspAddress);
         AddAction(0x62, GetPspAddress);
+        AddAction(0x63, GetLeadByteTable);
+    }
+
+    private void GetLeadByteTable() {
+        if (LoggerService.IsEnabled(LogEventLevel.Verbose)) {
+            LoggerService.Verbose("GET LEAD BYTE TABLE");
+        }
+        State.ES = 0;
+        State.BX = 0;
     }
 
     /// <summary>
--- a/src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
+++ b/src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
@@ -111,12 +111,16 @@ public class DosInt21Handler : InterruptHandler {
         AddAction(0x63, GetLeadByteTable);
     }
 
+    /// <summary>
+    /// Get a pointer to the "lead byte" table, for foreign character sets.
+    /// This is a table that tells DOS which bytes are the first byte of a double-byte character.
+    /// We don't support double-byte characters (yet), so we just return 0.
+    /// </summary>
     private void GetLeadByteTable() {
         if (LoggerService.IsEnabled(LogEventLevel.Verbose)) {
             LoggerService.Verbose("GET LEAD BYTE TABLE");
         }
-        State.ES = 0;
-        State.BX = 0;
+        State.AX = 0;
     }
 
     /// <summary>
--- a/src/Spice86.Core/Emulator/VM/EmulationLoop.cs
+++ b/src/Spice86.Core/Emulator/VM/EmulationLoop.cs
@@ -86,8 +86,8 @@ public class EmulationLoop {
     private void RunLoop() {
         _stopwatch.Start();
         while (_cpuState.IsRunning) {
-            _pauseHandler.WaitIfPaused();
             _machineBreakpoints.CheckBreakPoint();
+            _pauseHandler.WaitIfPaused();
             _cpu.ExecuteNextInstruction();
             //_cfgCpu.ExecuteNext();
             _timer.Tick();
--- a/src/Spice86.Core/Emulator/VM/EmulationLoop.cs
+++ b/src/Spice86.Core/Emulator/VM/EmulationLoop.cs
@@ -23,6 +23,7 @@ public class EmulationLoop {
     private readonly MachineBreakpoints _machineBreakpoints;
     private readonly DmaController _dmaController;
     private readonly Stopwatch _stopwatch;
+    private readonly IPauseHandler _pauseHandler;
 
     /// <summary>
     /// Initializes a new instance.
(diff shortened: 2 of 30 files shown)


src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -108,6 +108,15 @@ public class DosInt21Handler : InterruptHandler {
         AddAction(0x4F, () => FindNextMatchingFile(true));
         AddAction(0x51, GetPspAddress);
         AddAction(0x62, GetPspAddress);
+        AddAction(0x63, GetLeadByteTable);
+    }
+
+    private void GetLeadByteTable() {
```
maximilien-noal, line 114: What is this table ? What is its purpose ? Is it entirely implemented or os this a stub ? Why do we return 0 ?

Can all of this be documented in XML summary for this method ?

Also this method should be public, for overrides.

src/Spice86.Core/Emulator/VM/IPauseHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// Event triggered when a pause is requested on the emulator.
+    /// This allows other parts of the application to react to pauses,
+    /// such as stopping processing or updating UI elements.
+    /// </summary>
+    event EventHandler<EventArgs>? Pausing;
+
+    /// <summary>
+    /// Event triggered when a resume is requested on the emulator.
+    /// </summary>
+    event EventHandler<EventArgs>? Resumed;
```
maximilien-noal, line 22: If the source object parameter isn't important, this could be an action<T> instead of an EventHandler<T>

EventHandler were made for code generators.

src/Spice86.Core/Emulator/VM/IPauseHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// Gets a value indicating whether the emulator is currently paused.
+    /// </summary>
+    bool IsPaused { get; }
+
+    /// <summary>
+    /// Event triggered when a pause is requested on the emulator.
+    /// This allows other parts of the application to react to pauses,
+    /// such as stopping processing or updating UI elements.
+    /// </summary>
+    event EventHandler<EventArgs>? Pausing;
```
maximilien-noal, line 17: If the source object parameter isn't important, this could be an action instead of an EventHandler

EventHandler were made for code generators.

maximilien-noal (changes requested):

maximilien-noal:
LGTM. i only requested a documentation change.

I still need to properly test this tomorrow night.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
kevinferrare: Looks like this http://www.ctyme.com/intr/rb-3143.htm
More details here: http://www.ctyme.com/intr/rb-3142.htm#Table1746

Not sure why ES:BX is set while spec says DS:SI

src/Spice86.Core/Emulator/VM/EmulationLoop.cs
```diff
@@ -99,7 +86,7 @@ public class EmulationLoop {
     private void RunLoop() {
         _stopwatch.Start();
         while (_cpuState.IsRunning) {
-            PauseIfAskedTo();
+            _pauseHandler.WaitIfPaused();
```
kevinferrare, line 89: Just noticed WaitIfPaused is called here but also in _machineBreakpoints.CheckBreakPoint()
This is not due to your change, but while you are at it, I wonder if this is not someting that could be removed either here on in the MachineBreakpoints class.

maximilien-noal (dismissed):

maximilien-noal:
I just noticed that the UI will freeze if Dune is launched:

image

I can't show performance information for example.

Debug mode, command line args:

    -d false -s -a "ADP220 SBP2227" -e "/home/max/Jeux/Dune/DNCDPRG.EXE"

On master, I don't get this behavior.

maximilien-noal (changes requested):
Dune hangs the UI if run (see comment with the details)

JorisVanEijden:
> I just noticed that the UI will freeze if Dune is launched:
> 
> Debug mode, command line args:
> 
> ```
> -d false -s -a "ADP220 SBP2227" -e "/home/max/Jeux/Dune/DNCDPRG.EXE"
> ```
> 
> On master, I don't get this behavior.

I can't reproduce :(

On a Debug build with `-d false -s -a "ADP220 SBP2227" -e "C:\Games\DuneCD\Dncdprg.exe"`:

image

maximilien-noal:
> I can't reproduce :(

On my Xubuntu linux laptop, it works only if I don't attach the .NET Debugger.

In other words, a regular debug session is not possible.

I'll just assume it's a problem on my end and approve this PR. I saw that it works; just not with a .NET debugger attached.

maximilien-noal (approved):

maximilien-noal (changes requested):
Confirmed UI freeze on my Linux machine and on Kevin's machine.

Since it can't be reproduced everywhere, I'll try and see what caused it.

maximilien-noal (approved):

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
JorisVanEijden: I added docs and implemented a dummy 0 return.

We'll pick this up further when we start supporting asian character sets in DOS

src/Spice86.Core/Emulator/VM/IPauseHandler.cs
JorisVanEijden: Replaced with Action

src/Spice86.Core/Emulator/VM/EmulationLoop.cs
JorisVanEijden: I dropped the one from CheckBreakPoint and reorderd the calls so that if checking the breakpoint causes a pause, it is triggered immediately

JorisVanEijden:
Debugger issue traced to a thread sync issue with the linux thread scheduler. Worked around by adding a `Thread.Sleep(0)`

PR #794, opened 2024-07-29
Fix: dashes in passed cmdline args

### LINKED ISSUE #526: CommandLineParser gets confused by dashes inside the arguments command line string argument (it erases other existing arguments)
TODO: replace it with System.CommandLine

https://github.com/dotmake-build/command-line

maximilien-noal:
Sucessfully tesed with Doom8088.

Fixes #526

### DIFF
--- a/src/Spice86.Core/CLI/CommandLineParser.cs
+++ b/src/Spice86.Core/CLI/CommandLineParser.cs
@@ -9,6 +9,7 @@ using Spice86.Shared.Utils;
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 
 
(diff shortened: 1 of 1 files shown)


src/Spice86.Core/CLI/CommandLineParser.cs
```diff
@@ -31,6 +46,21 @@ public class CommandLineParser : ICommandLineParser {
             throw exception;
         });
     }
+    
+    private static string? GetCommandLineName<T>(string propertyName, bool shortName) {
```
kevinferrare, line 50: get -> generate? From what I understand this regenerates the name that could have been passed to the command line for the property?

src/Spice86.Core/CLI/CommandLineParser.cs
```diff
@@ -9,19 +9,34 @@
...
+using System.Linq;
 using System.Reflection;
 
 
 /// <inheritdoc cref="ICommandLineParser" />
 public class CommandLineParser : ICommandLineParser {
     /// <inheritdoc />
     public Configuration ParseCommandLine(string[] args) {
-        ParserResult<Configuration> result = Parser.Default.ParseArguments<Configuration>(args);
+        int indexOfPassedExeArgs = Array.IndexOf(args, GetCommandLineName<Configuration>(nameof(Configuration.ExeArgs), false));
```
kevinferrare, line 20: How about moving that to a method like ExtractDosExeArgs so that it's more obvious what this does?

src/Spice86.Core/CLI/CommandLineParser.cs
maximilien-noal: It gets them from the attribute via reflection.

It's an alternative to copying them in this file as '-a' or '-ExeArgs', as it might change.

src/Spice86.Core/CLI/CommandLineParser.cs
JorisVanEijden: or a slightly simpler method:
```csharp
    private static List<string> ProcessArgs(string[] args) {
        var processedArgs = new List<string>();
        for (int i = 0; i < args.Length; i++) {
            if (IsExeArg(args[i]) && i + 1 < args.Length) {
                processedArgs.Add(args[i]);
                processedArgs.Add(args[i + 1]);
                i++;
            } else {
                processedArgs.Add(args[i]);
            }
        }

        return processedArgs;
    }
```

src/Spice86.Core/CLI/CommandLineParser.cs
JorisVanEijden: It could use some comments to that effect.

src/Spice86.Core/CLI/CommandLineParser.cs
maximilien-noal: Done.

src/Spice86.Core/CLI/CommandLineParser.cs
maximilien-noal: Comment added.

kevinferrare (approved):

PR #808, opened 2024-08-22
New ASM test to check if external timer interrupts are handled properly

kevinferrare:


maximilien-noal:
A very important unit test that was missing! Looks good to me. :)

maximilien-noal (approved):

PR #855, opened 2024-09-18
Fix CfgCpu messing up graph on external interruptions. Make interrupt vector table a memory based array.

kevinferrare:
No dune crash with cfgcpu with this pr


maximilien-noal (approved):
LGTM. Will give it a spin with a Dune longplay this week-end. 😊

PR #871, opened 2024-09-26
feature: DOS LOAD AND/OR EXEC and DOS process management

maximilien-noal:
EDIT: For context, the previous description was essentially 'refactor: make the DOS INT21H load and execute the initial program', and the PR was limited to only just that. It isn't anymore, as it wasn't enough.

### Description of Changes

Implements LOAD AND/OR EXEC (DOS INT21H function 0x4B) and DOS process management.

### Rationale behind Changes

Makes LOAD AND/OR EXEC, used by a lot of DOS games, work. Also a lot of critical DOS kernel structures are present and initialized in emulated memory now, which at least makes our DOS implementation more like an expected one.

### Suggested Testing Steps

Test this with your favorite game.
It should still load, even if you set a preferred start segment in the config, or even use program arguments.

Dune still works.


JorisVanEijden:
Krondor loads and runs.

But I don't think an interrupt handler is the right place for this file loading is it?
Feels like `DosFileManager` would be a better fit.

I'm also not quite sure how this makes it easier to implement DOS function INT21H 0x4B.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -1069,4 +1074,105 @@ private void SetStateFromDosFileOperationResult(bool calledFromVm, DosFileOperat
...
+        SetEntryPoint(startSegment, ComOffset);
+        State.InterruptFlag = true;
+    }
+
+    internal void LoadHostFile(string hostFile, string? arguments, ushort startSegment, bool isCom) {
+        if (isCom) {
+            LoadHostFileAsCom(hostFile, arguments, startSegment);
+        } else {
+            LoadHostFileAsExe(hostFile, arguments, startSegment);
+        }
```
JorisVanEijden, line 1176: Not a bug fan of this construction.

The caller has to know if it's a com or exe, so it can just as well call the appropriate method itself.

maximilien-noal:
Dosbox does this :

Call the command processor, which eventually calls dos int21h 0x4b to execute a DOS program with a DOS path and args as parameters given via CPU registers.

It also handles the SDA structure, PSP chains, DOS process management, TSRs and such. 

What we do currently on master:

We completely side step the dos kernel and dos int handlers and the usage of a DOS path.

We directly load an executable in memory from an absolute host path, with class that are foreign to the OperatingSystem namespace. 

We don't have the SDA, DOS process management, TSR support, etc.

What this does:

What I see as the first step : call the dos int21h class to load the DOS program, from an absolute host path.

It's a little bit better, because the rest of the modifications will be done in the dos kernel and the dos int21h handler.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -776,13 +776,13 @@ public DosFileOperationResult RemoveDirectory(string dosDirectory) {
 
             return DosFileOperationResult.NoValue();
         } catch (IOException e) {
+            triedToDeleteCurrentDir = true;
```
JorisVanEijden, line 779: Is deleting the currentDir really the only cause for IOExceptions when deleting a directory?

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -776,13 +776,13 @@ public DosFileOperationResult RemoveDirectory(string dosDirectory) {
...
             if (_loggerService.IsEnabled(LogEventLevel.Warning)) {
-                _loggerService.Warning(e, "Error while deleting directory {CaseInsensitivePath}: {Exception}",
+                _loggerService.Warning(e, "Error while deleting directory {FullHostPath}: {Exception}",
                     fullHostPath, e);
             }
         }
-
-        return PathNotFoundError(dosDirectory);
+        return triedToDeleteCurrentDir ? RemoveCurrentDirError(dosDirectory) : PathNotFoundError(dosDirectory);
```
JorisVanEijden, line 785: Why not return this at the end of the codeblock where that boolean is set?

src/Spice86.Core/Emulator/ProgramExecutor.cs
```diff
@@ -47,46 +46,52 @@ public sealed class ProgramExecutor : IDisposable {
...
+        if (string.IsNullOrWhiteSpace(_configuration.Exe)) {
+            throw new UnrecoverableException("No file to execute");
+        }
+        string fileExtension = Path.GetExtension(_configuration.Exe).ToLowerInvariant();
+        if(fileExtension is ".com" or ".exe") {
+            _dosInt21Handler.LoadHostFile(_configuration.Exe, _configuration.ExeArgs, _configuration.ProgramEntryPointSegment,
+                isCom: fileExtension is ".com");
+        } else {
+            new BiosLoader(_memory, _state, _configuration.Exe, _loggerService).LoadHostFile();
+        }
```
JorisVanEijden, line 94: This might read better as a switch instead of nested if statements.

JorisVanEijden:
> Dosbox does this :
> 
> Call the command processor, which eventually calls dos int21h 0x4b to execute a DOS program with a DOS path and args as parameters given via CPU registers.
> 
> It also handles the SDA structure, PSP chains, DOS process management, TSRs and such.
> 
> What we do currently on master:
> 
> We completely side step the dos kernel and dos int handlers and the usage of a DOS path.
> 
> We directly load an executable in memory from an absolute host path, with class that are foreign to the OperatingSystem namespace.
> 
> We don't have the SDA, DOS process management, TSR support, etc.
> 
> What this does:
> 
> What I see as the first step : call the dos int21h class to load the DOS program, from an absolute host path.
> 
> It's a little bit better, because the rest of the modifications will be done in the dos kernel and the dos int21h handler.

The way I imagine it, we'd have a "FileLoader" class (or set of classes) that are used by both the int21handler and the ProgramExecutor.
I dunno, to me it just feels more like it belongs somewhere under the `Spice86.Core.Emulator.OperatingSystem` namespace rather than under the `Spice86.Core.Emulator.InterruptHandlers.Dos` namespace.

maximilien-noal:
I can do that.

maximilien-noal:
Closed, to be reworked in another form.

PR #874, opened 2024-10-01
feature: Internal debugger memory and execution breakpoints

maximilien-noal:
Introduces Breakpoint support in the DISASM view (Execution Breakpoints), Memory view (Memory Access/Read/Write breakpoint), and in a dedicated Breakpoints view.

Breakpoints created via the internal debugger can't go out of sync with the GDB support, since the internal debugger is not accessible if the GDB server is present.

image

image


src/Spice86/ViewModels/DisassemblyViewModel.cs
```diff
@@ -154,6 +158,23 @@ private static List<CpuInstructionInfo> DecodeInstructions(State state, IMemory
...
+        _pauseHandler.RequestPause($"breakpoint {breakPoint.BreakPointType} hit");
+
+    [RelayCommand]
+    private void CreateUnconditionalBreakpoint() {
+        if (SelectedInstruction is null) {
+            return;
+        }
+        AddressBreakPoint breakPoint = new(BreakPointType.EXECUTION, SelectedInstruction.Address, OnBreakPointReached, false);
+        _emulatorBreakpointsManager.ToggleBreakPoint(breakPoint, true);
+        SelectedInstruction.HasBreakpoint = _emulatorBreakpointsManager.IsAddressBreakpointAt(SelectedInstruction.Address);
```
kevinferrare, line 176: Since you just created a breakpoint at this address, why not just true?

maximilien-noal: Corrected.

src/Spice86/Models/Debugging/CpuInstructionInfo.cs
```diff
@@ -4,8 +4,12 @@ namespace Spice86.Models.Debugging;
 
 using Iced.Intel;
 
+using Spice86.Core.Emulator.VM.Breakpoint;
+
 public partial class CpuInstructionInfo : ObservableObject {
     [ObservableProperty] private string? _stringRepresentation;
+    [ObservableProperty] private bool _hasBreakpoint;
+    [ObservableProperty] private BreakPoint? _breakPoint;
```
JorisVanEijden, line 12: Do we need both?
Can they get out of sync?

maximilien-noal: We don't need both, I'll replace this entire pr with a dedicated BreakPoint management view and this ViewModel will only communicate with a IAddressBreakPointViewModel interface.

src/Spice86/Views/CpuStackView.axaml
```diff
@@ -0,0 +1,40 @@
...
+				IsVisible="{Binding IsPaused}">
+				<TextElement.FontFamily>
+					<OnPlatform Default="{StaticResource RobotoMonoFont}" />
+				</TextElement.FontFamily>
+				<avaloniaHex:HexEditor.Columns>
+					<rendering:OffsetColumn />
+					<rendering:HexColumn />
+					<rendering:AsciiColumn />
+				</avaloniaHex:HexEditor.Columns>
+			</avaloniaHex:HexEditor>
```
JorisVanEijden, line 37: Can you add the Copy/Show in structureview context menu here too?

src/Spice86/ViewModels/CpuViewModel.cs
```diff
@@ -21,12 +22,16 @@ public partial class CpuViewModel : ViewModelBase {
...
+    [ObservableProperty]
+    private CpuStackViewModel _stack;
+
+    public CpuViewModel(State state, Stack stack, IMemory memory, IPauseHandler pauseHandler, IUIDispatcher uiDispatcher) {
         _cpuState = state;
         pauseHandler.Pausing += () => uiDispatcher.Post(() => _isPaused = true);
         _isPaused = pauseHandler.IsPaused;
         pauseHandler.Resumed += () => uiDispatcher.Post(() => _isPaused = false);
         DispatcherTimerStarter.StartNewDispatcherTimer(TimeSpan.FromMilliseconds(400), DispatcherPriority.Normal, UpdateValues);
+        Stack = new CpuStackViewModel(stack, memory, pauseHandler, uiDispatcher);
```
JorisVanEijden, line 34: I'm not sure what the reason behind adding the stack-view to the CPU window is.
Would it not be easier to add an "Open Stack View" button to the stack register? Having it in a separate window is useful to observe changes when stepping through code.

maximilien-noal: Will do.

src/Spice86/Views/CpuStackView.axaml
maximilien-noal: Sure. In order to not duplicate XAML code, I think this will CpuStackViewModel will have both a MemoryViewModel, and the CpuStackViewModel itself will serve this kind of view in a first TabControl:

373889212-d0933287-d24c-4fbe-8cfc-947fd851fbcb

See the origin of the idea here: https://github.com/epasveer/seer/issues/127

JorisVanEijden: sounds good

src/Spice86/ViewModels/CpuViewModel.cs
maximilien-noal: CPU Stack View removed. I will do it for a later PR.

JorisVanEijden (approved):

PR #905, opened 2024-10-26
chore: Nuget packages update

maximilien-noal:


src/Directory.Build.props
```diff
@@ -0,0 +1,77 @@
+<Project>
+	<!-- Common packages -->
```
JorisVanEijden, line 2: I don't think these are all required by this project.

src/Bufdio.Spice86/Bufdio.Spice86.csproj
```diff
@@ -1,56 +1,21 @@
...
-		<version>7.0.0</version>
-		<Authors>Luthfi Tri Atmaja, Kevin Ferrare, Maximilien Noal, Joris van Eijden, Artjom Vejsel</Authors>
-		<PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
-		<Description>Reverse engineer and rewrite real mode dos programs</Description>
-		<PackageProjectUrl>https://github.com/OpenRakis/Spice86</PackageProjectUrl>
-		<RepositoryUrl>https://github.com/OpenRakis/Spice86</RepositoryUrl>
-		<RepositoryType>git</RepositoryType>
+		<TargetFramework>net8.0</TargetFramework>
+		<Nullable>enable</Nullable>
+		<ImplicitUsings>enable</ImplicitUsings>
```
JorisVanEijden, line 6: This means some of the usings in the source files can be removed

src/Directory.Build.props
maximilien-noal: No they are not. But since dependabot is still broken, at least that way it will only mess up one file.

JorisVanEijden: ah, the GitHub interface made it look like this file was in the Bufdio.Spice86 directory.

But I don't feel this is the right way to solve issues with dependabot.
All of the projects now have unused dependencies...

We can look into Dependabot alternatives (at work I use Renovate). Or we can just manually click the "Upgrade packages in solution" button in Rider once in a while.

maximilien-noal: I'll put non common packages werr they belong so we can move on with this PR. 😊

Yeah replacing dependabot sounds good.

JorisVanEijden: https://github.com/apps/renovate

maximilien-noal: I re-did the PR, now Directory.Build.props only contains common packages (Sourcelink, source code analyzers)

src/Bufdio.Spice86/Bufdio.Spice86.csproj
maximilien-noal: Done.

JorisVanEijden (approved):

PR #906, opened 2024-10-26
Sync minor additions

JorisVanEijden:
### Description of Changes
- Add Krondor image cause it looks nice.
- Add argument fetcher and DoOnMemoryRead breakpoint helper
- Add verbose logging for when VGA switches to a different lbokc of memory to display.

### Rationale behind Changes
Just some tiny things I added along the way and that are annoying to redo when testing Spice86 MRs

### Suggested Testing Steps
...


src/Spice86.Core/Emulator/ReverseEngineer/ArgumentFetcher.cs
```diff
@@ -96,6 +96,11 @@ public void Get(out uint arg1, out uint arg2, out ushort arg3) {
         arg3 = _stack.Peek16(12);
     }
 
+    public void Get(out uint arg1, out uint arg2) {
```
maximilien-noal, line 99: Why not return an ValueTuple instead of void + out parameters ?

Especially since it only returns data.

maximilien-noal (changes requested):
Looks good to me. The only change I'd request is to return something instead of void + out parameters.

src/Spice86.Core/Emulator/ReverseEngineer/ArgumentFetcher.cs
JorisVanEijden: Because you cannot overload on return values only. The arguments are used to determine which Get() method to use.

And even if you could,
```csharp
ArgumentFetcher.StringShortUshort args = _args.Get(); 
string dosPath = args.Arg1;
short flags = args.Arg2;
ushort mode = args.Arg3;
```
and
```csharp
public record struct StringShortUshort(string Arg1, short Arg2, ushort Arg3);
public StringShortUshort Get() {
    return new StringShortUshort(
        GetStringFromDsPointer(_stack.Peek16(4)),
        (short)_stack.Peek16(6),
        _stack.Peek16(8)
    );
}
```

are not clearer than 

```csharp
_args.Get(out string dosPath, out short flags, out ushort mode);
```
and
```
public void Get(out string arg1, out short arg2, out ushort arg3) {
    ushort stringPointerOffset = _stack.Peek16(4);
    arg2 = (short)_stack.Peek16(6);
    arg3 = _stack.Peek16(8);
    arg1 = GetStringFromDsPointer(stringPointerOffset);
}
```

maximilien-noal: Oh, I see. I didn't see the other Get methods.

maximilien-noal (approved):

kevinferrare (approved):

PR #926, opened 2024-11-09
Fix/performance measurer high samples

maximilien-noal:
### Description of Changes

Fixes performances measurements which were incorrectly measured as 0 if samples were given with a time resolution below one millisecond to the PerformanceMeasurer class.

Also makes the EmulationLoop uses the PerformanceMeasurer to record and calculate performance measures, instead of recreating the function by itself.

### Rationale behind Changes

Just a small fix I came across.

### Suggested Testing Steps

Already tested. Nothing changed.

### DIFF
--- a/src/Spice86.Shared/Diagnostics/PerformanceMeasurer.cs
+++ b/src/Spice86.Shared/Diagnostics/PerformanceMeasurer.cs
@@ -44,7 +44,7 @@ public class PerformanceMeasurer : IPerformanceMeasurer {
         _lastTimeInMilliseconds = newTimeInMilliseconds;
         long valueDelta = newMeasure - _measure;
         _measure = newMeasure;
-        ValuePerMillisecond = valueDelta / Math.Max(millisecondsDelta, 1);
+        ValuePerMillisecond = valueDelta / millisecondsDelta;
         AverageValuePerSecond = ApproxRollingAverage(AverageValuePerSecond, ValuePerSecond, _sampledMetricsCount++);
     }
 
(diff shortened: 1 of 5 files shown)


src/Spice86.Shared/Diagnostics/PerformanceMeasurer.cs
```diff
@@ -31,21 +31,36 @@ public class PerformanceMeasurer : IPerformanceMeasurer {
...
         }
+
         long millisecondsDelta = newTimeInMilliseconds - _lastTimeInMilliseconds;
+        if (millisecondsDelta == 0) {
+            return;
+        }
         _lastTimeInMilliseconds = newTimeInMilliseconds;
         long valueDelta = newMeasure - _measure;
         _measure = newMeasure;
         ValuePerMillisecond = valueDelta / Math.Max(millisecondsDelta, 1);
```
JorisVanEijden, line 47: Since you added a check for 0 above, you can get rid of the Max()
```suggestion
        ValuePerMillisecond = valueDelta / millisecondsDelta;
```

src/Spice86/ViewModels/MainWindowViewModel.cs
```diff
@@ -228,14 +235,15 @@ public void SetResolution(int width, int height) {
...
-        //TODO: refactor this
         await _hostStorageProvider.DumpEmulatorStateToFile().ConfigureAwait(false);
     }
 
     [RelayCommand(CanExecute = nameof(CanPause))]
     public void Pause() => _pauseHandler.RequestPause("Pause button pressed in main window");
 
-    private void SetMainTitle() => MainTitle = $"{nameof(Spice86)} {Configuration.Exe}";
+    private void SetMainTitle(double instructionsPerMillisecond) {
+        MainTitle = $"{nameof(Spice86)} {Configuration.Exe} - {instructionsPerMillisecond:N0} cycles/ms";
```
JorisVanEijden, line 245: Slightly less jerky:
```suggestion
        MainTitle = $"{nameof(Spice86)} {Configuration.Exe} - cycles/ms: {instructionsPerMillisecond,7:N0}";
```

JorisVanEijden (approved):
Just 2 minor suggestions.

PR #948, opened 2024-11-26
Feat/debugger QoL

maximilien-noal:
### Description of Changes

Introduces keyboard shortcuts and tooltips everywhere, or almost everywhere. Including the Breakpoints view.

Makes Step Into 'go to CS:IP' only if needed. Meaning we can see what we already executed, unless CS:IP is out of viewable range. In this case we refresh the disassembly, and the instruction at CS:IP becomes the first line of the disassembly listing again.

### Rationale behind Changes

Makes the debugger behave more like the one found in Visual Studio.

Makes Step into more readable.

### Suggested Testing Steps

Already tested. But tests and suggestions are welcome!


src/Spice86/ViewModels/DisassemblyViewModel.cs
```diff
@@ -21,6 +21,7 @@ namespace Spice86.ViewModels;
 using Spice86.Shared.Utils;
 
 using System.Globalization;
+using System.Net.NetworkInformation;
```
JorisVanEijden, line 24: Don't need this do we?

JorisVanEijden (approved):
Other than the unused using this looks fine.

PR #998, opened 2024-12-28
Internal Debugger: All breakpoints can be created (including I/O ports breakpoints) and edited + CPU Stack View

maximilien-noal:
### All breakpoints types are supported

I/O Port, Cycles, CPU execution address, Memory access/write/read, and interrupt; all are supported by the internal debugger.

Creation and editing, not possible before, has been implemented.

For creating an interrupt breakpoint, it can be done with (for example) either '0x21' or '21H'

## QoL

The overall presentation has been improved, most notably the dissassembly view:

image

Also, hexadecimal addresses and values are more often used. And presented with 0x before them.

### Segmented Addresses

Segmented Addresses are supported when creating a breakpoint in the DISASM view, and in the Breakpoints Manager.

Expressions such as CS:IP are resolved at time of creation, before creating the breakpoint (once).

## CPU Stack Memory

Adds a CPU Stack Memory View to the Memory tab in the Internal Debugger.

It reuses the current memory view and MemoryViewModel.

image


src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -67,6 +67,10 @@ public DosInt21Handler(IMemory memory, Cpu cpu,
         FillDispatchTable();
     }
 
+    public override void Run(int index) {
+        base.Run(index);
+    }
+
```
JorisVanEijden, line 73: What does this do?

src/Spice86/UserControls/AddressAutoCompleteBox.axaml.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+
+                if (split.Length > 1 &&
+                    ushort.TryParse(firstPart, NumberStyles.Integer,
+                    CultureInfo.InvariantCulture, out ushort segment) &&
+                    ushort.TryParse(secondPart, NumberStyles.Integer,
+                    CultureInfo.InvariantCulture, out ushort offset)) {
+                    address = Math.Min(A20Gate.EndOfHighMemoryArea, MemoryUtils.
+                        ToPhysicalAddress(segment, offset));
+                    return true;
```
JorisVanEijden, line 58: I don't understand.  Any input that passes this surely also passed the previous check, so this code will never do anything?

src/Spice86/UserControls/AddressAutoCompleteBox.axaml.cs
```diff
@@ -0,0 +1,128 @@
...
+    private string ReplaceRegisterNameWithValue(string part, State state) {
+        return part.ToUpperInvariant() switch {
+            "CS" => state.CS.ToString("X4"),
+            "IP" => state.IP.ToString("X4"),
+            "DS" => state.DS.ToString("X4"),
+            "ES" => state.ES.ToString("X4"),
+            "FS" => state.FS.ToString("X4"),
+            "GS" => state.GS.ToString("X4"),
+            "SS" => state.SS.ToString("X4"),
+            _ => part
```
JorisVanEijden, line 85: No love for the other registers?
`ES:DI`, `DS:SI` and `AX:DX` could be fairly common usecases I think.

src/Spice86/UserControls/AddressAutoCompleteBox.axaml.cs
```diff
@@ -0,0 +1,128 @@
...
+using Spice86.Core.Emulator.CPU;
+using Spice86.Core.Emulator.Memory;
+using Spice86.Shared.Utils;
+
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+
+namespace Spice86.UserControls;
+
+public partial class AddressAutoCompleteBox : UserControl {
```
JorisVanEijden, line 13: This doesn't actualy autocomplete anything. `AddressAutoConversionBox` might be a better name?

src/Spice86/ViewModels/BreakpointViewModel.cs
```diff
@@ -2,33 +2,38 @@ namespace Spice86.ViewModels;
...
-            string comment = "") {
+        long addressOrIoPortOrCyclesOrInterruptNumber,
+        BreakPointType type,
+        bool isRemovedOnTrigger,
+        Action onReached,
+        string comment = "")
+    {
         _emulatorBreakpointsManager = emulatorBreakpointsManager;
-        Address = address;
+        Address = addressOrIoPortOrCyclesOrInterruptNumber;
```
JorisVanEijden, line 24: Maybe `trigger` is a friendlier name?

src/Spice86/ViewModels/BreakpointsViewModel.cs
```diff
@@ -1,61 +1,294 @@
...
+        new TabItemViewModel { Header = "Interrupt", IsSelected = false },
+        new TabItemViewModel { Header = "I/O Port", IsSelected = false }
+    };
+
+    private TabItemViewModel _selectedBreakpointTypeTab;
+
+    public TabItemViewModel? SelectedBreakpointTypeTab {
+        get => _selectedBreakpointTypeTab;
+        set {
+            if(value is not null) {
```
JorisVanEijden, line 61: What should happen if it _is_ `null`?

src/Spice86/ViewModels/BreakpointsViewModel.cs
```diff
@@ -1,61 +1,294 @@
...
+                OnPropertyChanged(nameof(IsCyclesBreakpointSelected));
+                OnPropertyChanged(nameof(IsInterruptBreakpointSelected));
+                OnPropertyChanged(nameof(IsIoPortBreakpointSelected));
+            }
+        }
+    }
+
+    public State State => _state;
+
+    public bool IsExecutionBreakpointSelected => BreakpointTabs[2].IsSelected;
```
JorisVanEijden, line 77: This feels very error-prone, depending on the number and order of tab items.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
maximilien-noal: No changes compared to master. This is a false positive.

src/Spice86/UserControls/AddressAutoCompleteBox.axaml.cs
maximilien-noal: File removed.

src/Spice86/UserControls/AddressAutoCompleteBox.axaml.cs
maximilien-noal: File removed.

src/Spice86/UserControls/AddressAutoCompleteBox.axaml.cs
maximilien-noal: File removed.

src/Spice86/ViewModels/BreakpointsViewModel.cs
maximilien-noal: If it is null, I don't do anything.

src/Spice86/ViewModels/BreakpointViewModel.cs
```diff
@@ -54,7 +80,7 @@ public bool IsEnabled {
 
     public bool IsRemovedOnTrigger { get; }
 
-    public uint Address { get; }
+    public long Address { get; }
```
JorisVanEijden, line 83: This is a breaking change I did not expect. Do we have a usecase for addresses below 0 or above 0xFFFFFFFF ?

maximilien-noal: It represent a trigger : an I/O port number, a cycles number, a memory address, or an interrupt number.

Guess it could be int.

JorisVanEijden (approved):
Yeah, let's get this merged.

PR #1031, opened 2025-02-04
Add IO and Interrupt breakpoints

kevinferrare:
### Description of Changes

Add breakpoints for port read / write (same as for memory) and for Interrupt based on vector number


maximilien-noal (approved):
LGTM. Great work !

Can't wait to try it out with the future internal debugger UI improvments and features. :)

PR #1066, opened 2025-02-27
fix: Notification of pauses from emulator to GDB client

maximilien-noal:
This fixes notifications of pauses from the emulator to the gdb client.

That is, when one pauses the emulator from outside the GDB client (namely, the spice86 UI), GDB is notified.
This makes the GDB command prompt available again.

image


kevinferrare (approved):
Working very nicely :)

PR #1068, opened 2025-02-28
Add a minimum number of ticks between activations for PIT8254 counters

kevinferrare:
### Description of Changes
PIT8254 counters will not be able to activate unless 100 CPU cycles have passed since last activation for both timer based and cycle based implementations.
This was only done for timer based implementations before, and for performances reason.

### Rationale behind Changes
Some games like stunts re-enable interrupts in INT8 handler. If another interrupt happens before current int8 is finished, it will fill the call stack and cause a stack overflow, and game will crash.

### Suggested Testing Steps
Dune looks happy

### DIFF
--- a/src/Spice86.Core/Emulator/Devices/Timer/CounterActivator.cs
+++ b/src/Spice86.Core/Emulator/Devices/Timer/CounterActivator.cs
@@ -6,6 +6,7 @@ using Spice86.Core.Emulator.VM;
 /// Base class for counter activators
 /// </summary>
 public abstract class CounterActivator {
+    protected const int MinTicksBetweenActivation = 100;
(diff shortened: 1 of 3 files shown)


maximilien-noal (approved):

JorisVanEijden (changes requested):
As discussed on Discord. This is not really a solution, but a workaround for a hidden bug

maximilien-noal (changes requested):
Maybe this PR should be closed, since this was already fixed ?

kevinferrare:
Let's close it. Not sure on real hardware if you could lock the CPU by having the timer set so high that every instruction would get an interrupt. In any case, this would probably be intentional.

PR #1070, opened 2025-03-02
Fix PIT8254 and plug counter 2 into the PC Speaker device

kevinferrare:
### Description of Changes
- Make PIT8254 set counters on read write policy #3 (read / write the whole 16 bit value of the counter).
This seems to be the expected value for games that don't set it.

- Rework the way the counter code handles writes and latch.
Not sure if we should stop using the value until it has been fully written in mode 3, doc doesnt seem very explicit about this.
https://www.scs.stanford.edu/23wi-cs212/pintos/specs/8254.pdf

- Plugged the counter number 2 of the PIT into the PcSpeaker device so that it is aware of frequency changes

- Reworked the PC speaker with a thread so that some sound can be produced. Sound is a bit off.

### Rationale behind Changes
Changing default read/write policy ensures games that dont set it can get the frequency they expect for the PIT
Noticed that PC Speaker code was also half implementing a counter so factorized it

### Suggested Testing Steps
Working with dune and krondor


maximilien-noal (approved):
Great work ! It clears up a lot.

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
```diff
@@ -204,39 +185,45 @@ public override void WriteByte(ushort port, byte value) {
...
         }
 
-        switch (port)
-        {
-            case 0x61:
-            {
-                SpeakerControl oldValue = _controlRegister;
-                _controlRegister = (SpeakerControl)value;
-                if ((oldValue & SpeakerControl.SpeakerOn) != 0 && (_controlRegister & SpeakerControl.SpeakerOn) == 0) {
+        switch (port) {
```
JorisVanEijden, line 188: Perhaps the ReadByte and WriteByte should handle the port check in the same way?

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
```diff
@@ -71,42 +80,43 @@ public PcSpeaker(SoftwareMixer softwareMixer, State state, IOPortDispatcher ioPo
...
     }
-    
+
     /// <summary>
     /// Invoked when the frequency has changed.
     /// </summary>
     /// <param name="source">Source of the event.</param>
     /// <param name="e">Unused EventArgs instance.</param>
-    private void FrequencyChanged(object? source, EventArgs e) {
+    private void OnFrequencyChange(object? source, EventArgs e) {
```
JorisVanEijden, line 95:
```suggestion
    private void OnFrequencyChanged(object? source, EventArgs e) {
```

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
```diff
@@ -125,64 +135,35 @@ private int GenerateSquareWave(Span<byte> buffer, int period) {
...
+            if (!IsEnabled(_controlRegister)) {
+                continue;
+            }
+            int samples = GenerateSquareWave(buffer, _currentNote.Period);
+            int periods = _currentNote.PeriodCount;
 
-                while (_soundChannel.Render(floatArray.AsSpan()) > 0) {
-                    await Task.Yield();
-                }
+            // stereo
```
JorisVanEijden, line 151:
```suggestion
            // While the original PC speaker was mono, the emulator output is stereo. So we have to duplicate the signal.
```

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
```diff
@@ -204,39 +185,45 @@ public override void WriteByte(ushort port, byte value) {
...
+                SpeakerControl newValue = (SpeakerControl)value;
+                if (IsEnabled(_controlRegister) && !IsEnabled(newValue)) {
                     SpeakerDisabled();
+                } else {
+                    if (!_playbackStarted) {
+                        _loggerService.Information("Starting thread '{ThreadName}'", _playbackThread.Name);
+                        _playbackStarted = true;
+                        _playbackThread.Start();
+                    }
                 }
```
JorisVanEijden, line 199:
```suggestion
                } else if (!_playbackStarted) {
                    _loggerService.Information("Starting thread '{ThreadName}'", _playbackThread.Name);
                    _playbackStarted = true;
                    _playbackThread.Start();
                }
```

src/Spice86.Core/Emulator/Devices/Timer/Pit8254Counter.cs
```diff
@@ -0,0 +1,175 @@
...
+
+    /// <summary>
+    /// Gets the index of the counter.
+    /// </summary>
+    public int Index { get; init; }
+
+    /// <summary>
+    /// Gets or sets the mode of operation for the counter.
+    /// <br/>
+    /// Looks like mode 3 is the mode games expect to find when they start
```
JorisVanEijden, line 60: Is this true? I think both Dune and Krondor explicitly set this to 3.

AlbertoMarnetto: Seems that at least the BIOS of IBM PC XT and AT set it to 3:

https://github.com/gawlas/IBM-PC-BIOS/blob/e6cae33370fa7cd0568d72b785f682971edcc70c/IBM%20PC%20XT/XT%20BIOS%20V3/POST.ASM#L715

https://github.com/gawlas/IBM-PC-BIOS/blob/e6cae33370fa7cd0568d72b785f682971edcc70c/IBM%20PC%20AT/AT%20BIOS%20V3/TEST2.ASM#L284

The control register gets set in various ways during the boot, but these lines seem the last ones to touch it on the respective platforms.

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
kevinferrare: I guess you are talking about switch vs if.
It was a remain from the old pc speaker code that also implemented a very basic counter on port 42.
I uniformized both methods

src/Spice86.Core/Emulator/Devices/Timer/Pit8254Counter.cs
kevinferrare: Thanks for finding this, I checked the BIOS code but couldnt find that.
I added this reference to our code

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
kevinferrare: After looking deeper into this, it looks like most of the counter settings can affect sound.
I changed it to OnTimerSettingChanged and did the same for the event Counter side.

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
```diff
@@ -1,66 +1,75 @@
...
+        _playbackThread = new Thread(AudioPlayback) {
+            Name = nameof(PcSpeaker),
+        };
     }
 
     /// <summary>
     /// Gets the current frequency in Hz.
     /// </summary>
-    private double Frequency => FrequencyFactor / _frequencyRegister;
+    private double Frequency => _pit8254Counter.Activator.Frequency;
```
maximilien-noal, line 67: The class talks to 'a friend of a friend' here.

should it not directly depend on _activator.Frequency ?

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
```diff
@@ -204,39 +186,38 @@ public override void WriteByte(ushort port, byte value) {
             _loggerService.Verbose("PC Speaker set value {PCSpeakerValue}", ConvertUtils.ToHex8(value));
```
maximilien-noal, line 186: Could use a check if _loggerService.IsEnabled... for performance reasons.

src/Spice86.Core/Emulator/Devices/Sound/PCSpeaker/PcSpeaker.cs
```diff
@@ -204,39 +186,38 @@ public override void WriteByte(ushort port, byte value) {
...
-                    SpeakerDisabled();
-                }
-
-                break;
+        if (port == PcSpeakerPortNumber) {
+            SpeakerControl newValue = (SpeakerControl)value;
+            if (IsEnabled(_controlRegister) && !IsEnabled(newValue)) {
+                SpeakerDisabled();
+            } else if (!_playbackStarted) {
+                _loggerService.Information("Starting thread '{ThreadName}'", _playbackThread.Name);
```
maximilien-noal: Same here with _loggerService.IsEnabled

maximilien-noal (approved):

PR #1076, opened 2025-03-04
Use device thread instead of duplicating thread implementation for each device that needs a thread

kevinferrare:
### Description of Changes
Devices that require threads were handling them on their own with duplicated code (all are sound devices).

This expands the use of DeviceThread class used for PC speaker.

### Rationale behind Changes
More robust thread handling in devices
Unified logging for device start
Sound threads all pause when the emulator pauses

### Suggested Testing Steps
Tested:
- midi and mt32 with dune
- soundblaster with dune
- pc speaker with prince of persia and stunts
- OPL3FM with prince of persia and krondor


maximilien-noal:
Very nice work ! This saves up a lot of typing for a new sound device.

PR #1078, opened 2025-03-05
Fix pause events

JorisVanEijden:
### Description of Changes
Added "Paused" event for systems that want to react on the system finishing entering the paused state.
Moved UI systems from "Pausing" event (which signals the start of enetering the paused state) to "Paused".

### Rationale behind Changes
"Pausing" is for systems that need to do something before a pause can take full effect.
"Paused" is for systems that want to react to the system having become paused.
UI systems using Pausing could read values from the CPU while it was not yet paused. The debugger got confused when the values from the current cpu state did not match the values read when the Pausing event was handled.

### Suggested Testing Steps
Use the debugger to see if behaviour is consistent now.


maximilien-noal (approved):
LGTM. :) Thanks!

PR #1080, opened 2025-03-08
Call flow handler for CFG CPU

kevinferrare:
### Description of Changes
Brings call flow tracking to CFG CPU.

- This allows to track the function call stack whether in regular or cfgcpu.
- This also allows to link CALL instructions to the instructions their associated returns return to. If return is not at the expected place, a link of type CallToMisalignedReturn is created, otherwise it is CallToReturn.

Main changes:
- Introduction of function catalogue class to separate loading, listing and creating function names from the actual function tracking
- Introduction of new interfaces for CPU and cfgcpu to make them truly interchangeable
- Removal of hard references to CPU.cs class
- Modification of function tracking code (function handler) to work with cfgcpu
- Modification of cfgcpu execution context tracking code to make function handler work in separate contexts
- Tests can now access the generated graph in cfgcpu mode and check it is as expected

### Suggested Testing Steps
Tested only with Dune, Kronder, Stunts, POP.
GDB is still working properly.


src/Spice86.Core/Emulator/CPU/CfgCpu/Linker/NodeLinker.cs
```diff
@@ -89,18 +147,37 @@
...
     }
+
+    private void ReplaceSuccessorOfCallInstruction(CfgInstruction instruction, ICfgNode currentSuccesor, ICfgNode newSuccesor) {
+        foreach(KeyValuePair<InstructionSuccessorType, ISet<ICfgNode>> entry in instruction.SuccessorsPerType) {
+            ISet<ICfgNode> successors = entry.Value;
+            if (successors.Contains(currentSuccesor)) {
+                successors.Remove(currentSuccesor);
+                successors.Add(newSuccesor);
+            }
+        }
```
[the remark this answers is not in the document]
kevinferrare: No, NodeLinker is performance critical ...

src/Spice86.Core/Emulator/Function/FunctionCatalogue.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Spice86.Core.Emulator.Function;
+
+using Spice86.Shared.Emulator.Memory;
+
+public class FunctionCatalogue {
+    public FunctionCatalogue() : this(new Dictionary<SegmentedAddress, FunctionInformation>()) {
```
JorisVanEijden, line 6: So the same function can be in this dictionary with multiple SegmentedAddresses. Is that the way it should work or could that be problematic?

src/Spice86.Core/Emulator/CPU/CfgCpu/ParsedInstruction/InstructionField.cs
JorisVanEijden, line 1: XML comments need encoding:
`/// - 1234 would be a InstructionField&lt;ushort&gt;`

JorisVanEijden (approved):

src/Spice86.Core/Emulator/Function/FunctionCatalogue.cs
kevinferrare: That would be bad, would not make any sense. I guess you are thinking this could happen because we give a dictionary instead of a list? I will try to change it, this is an improvement

JorisVanEijden: I think I remember krondor having a function to convert a 32bit address to a seg:ofs combination. Probably because it loads large files and it's easier to calculate offsets that way.
Not sure if it also uses these to call functions, or only to access structs.

It also has a lot of indirect calls. Where the main code calls and address that has a jmp to the real function, wherever it is. But the library calls those functions directly. So functionally the same function, but with different addresses.

kevinferrare: If a function is called via a different segmented address, this implies wild things. Offsets calculations for near jumps could end up at different places like they do in ghidra, and code that accesses variables at CS:offset would access different values...

src/Spice86.Core/Emulator/CPU/CfgCpu/ExecutionContextManager.cs
```diff
@@ -1,61 +1,113 @@
...
+        _functionCatalogue = functionCatalogue;
+        // Initial fake but non-null context at init
+        CurrentExecutionContext = NewExecutionContext(SegmentedAddress.ZERO);
     }
 
+    private int CurrentDepth {
+        get => _currentDepth;
+        set {
+            _currentDepth = value;
+            _loggerService.LoggerPropertyBag.ContextIndex = value;
```
maximilien-noal, line 41: Is it possible to do `_loggerPropertyBag.ContextIndex = value;` instead ?

src/Spice86.Core/Emulator/CPU/CfgCpu/ExecutionContextReturns.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Spice86.Core.Emulator.CPU.CfgCpu;
+
+using Spice86.Shared.Emulator.Memory;
+using Spice86.Core.Emulator.CPU.CfgCpu.Linker;
+
+/// <summary>
+/// Maps return address to a stack of execution contexts to restore.
+/// Stack is because in some cases, code can call a new execution context that will return to the same address as the current one, so we need to know the call order to restore the correct one
+/// Cant be done with breakpoints since sometimes external int will happen even before the instruction after the iret is executed
```
maximilien-noal, line 9: I don't understand this comment line ? Can it be expanded upon ?

src/Spice86.Core/Emulator/Function/Dump/RecorderDataWriter.cs
```diff
@@ -36,19 +36,19 @@ public RecorderDataWriter(State state,
     /// <summary>
     /// Dumps all recorded data to their respective files.
     /// </summary>
-    public void DumpAll(ExecutionFlowRecorder executionFlowRecorder, FunctionHandler functionHandler) {
+    public void DumpAll(ExecutionFlowRecorder executionFlowRecorder, FunctionCatalogue functionCatalogue) {
         _loggerService.Verbose("Dumping all data to {DumpDirectory}", DumpDirectory);
```
maximilien-noal, line 40: Could use a check if Verbose level is enabled first

src/Spice86.Core/Emulator/CPU/IInstructionExecutor.cs
```diff
@@ -3,11 +3,21 @@ namespace Spice86.Core.Emulator.CPU;
...
 public interface IInstructionExecutor {
     /// <summary>
     /// Executes the next instruction.
     /// </summary>
     public void ExecuteNext();
+
+    /// <summary>
+    /// Signal that we are at the entry point of the program ready to start executing our very first instruction
+    /// </summary>
+    public void SignalEntry();
```
maximilien-noal, line 17: I wonder how this will work with DOS sub-processes, COMMAND.COM / batch files, and TSRs if it's implemented someday. Just a tought for the far future.

src/Spice86.Core/Emulator/CPU/CfgCpu/Linker/ExecutionContext.cs
```diff
@@ -1,13 +1,34 @@
...
+    /// </summary>
+    public SegmentedAddress EntryPoint { get; }
+    
+    /// <summary>
+    /// Function handler tracking the functions for this context. Function call stack is context dependant.
+    /// </summary>
+    public FunctionHandler FunctionHandler { get; }
+
+    /// <summary>
+    /// Depth at with this context was created.
```
maximilien-noal, line 30: Depth at which ?

src/Spice86.Core/Emulator/CPU/CfgCpu/Feeder/InstructionsFeeder.cs
```diff
@@ -37,15 +40,33 @@ public CfgInstruction GetInstructionFromMemory(SegmentedAddress address) {
...
+    }
+
+    private CfgInstruction ParseEnsuringUnique(SegmentedAddress address) {
         CfgInstruction parsed = _instructionParser.ParseInstructionAt(address);
-        _currentInstructions.SetAsCurrent(parsed);
-        _previousInstructions.AddInstructionInPrevious(parsed);
+        // Let's try with the discriminator reducer to see if the parsed instruction has an existing match
+        HashSet<CfgInstruction>? previousSet = PreviousInstructions.GetAtAddress(address);
+        if (previousSet != null) {
+            foreach (CfgInstruction existing in previousSet) {
```
maximilien-noal, line 63: could be a for loop instead of a foreach for more performance.

src/Spice86.Core/Emulator/CPU/CfgCpu/ControlFlowGraph/CfgNode.cs
```diff
@@ -14,6 +14,7 @@ public CfgNode(SegmentedAddress address) {
     public HashSet<ICfgNode> Predecessors { get; } = new();
     public HashSet<ICfgNode> Successors { get; } = new();
     public SegmentedAddress Address { get; }
+    public virtual bool CanCauseContextRestore { get; } = false;
```
maximilien-noal, line 17: false is the default value.

maximilien-noal (approved):
Some minor changes suggested in comments. LGTM !

src/Spice86.Core/Emulator/CPU/IInstructionExecutor.cs
kevinferrare: We could have a specific type of call for that maybe? Not sure how about the implications.
Anyway, this change is just to materialize that CPU needs to know that execution will start and stop, this is nothing new actually.

src/Spice86.Core/Emulator/CPU/CfgCpu/Feeder/InstructionsFeeder.cs
kevinferrare: When this piece of code is reached, parser has been called, reducer is being called in a loop.
This will probably not be noticeable since the overhead is way smaller than those 2 other systems.
Fortunately this is not called very often.

src/Spice86.Core/Emulator/CPU/CfgCpu/ExecutionContextReturns.cs
kevinferrare: Tried my best to explain it, topic is no walk in a park.

src/Spice86.Core/Emulator/CPU/CfgCpu/ExecutionContextManager.cs
kevinferrare: As discussed, we already depend on _loggerService for logging in this class so let's not add an additional property

src/Spice86.Core/Emulator/Function/Dump/RecorderDataWriter.cs
kevinferrare: Changed to information and included it in an if

PR #1082, opened 2025-03-09
Segmented address improvements

JorisVanEijden:
### Description of Changes
- It's now truly immutable and never needs to recalculate the linear address.
- The addition and subtraction operators got bug fixes for over and underflows.
- A bunch of other comparison operators were added.
- Some constructs that assumed SegmentedAddress was mutable or a reference type were fixed.

### Rationale behind Changes
If the SegmentedAddress is more flexible and easier to use, then perhaps it'll be easier to keep address segments for longer before converting them to uints

### Suggested Testing Steps
I ran Krondor and Dune for a bit and noticed no issues.

### DIFF
--- a/src/Spice86.Core/Emulator/InterruptHandlers/Common/MemoryWriter/MemoryAsmWriter.cs
+++ b/src/Spice86.Core/Emulator/InterruptHandlers/Common/MemoryWriter/MemoryAsmWriter.cs
@@ -30,7 +30,7 @@ public class MemoryAsmWriter : MemoryWriter {
     /// <param name="callbackNumber">Callback index</param>
     /// <param name="runnable">Action to run when this callback is executed by the CPU</param>
     public void RegisterAndWriteCallback(byte callbackNumber, Action runnable) {
-        Callback callback = new Callback(callbackNumber, runnable, GetCurrentAddressCopy());
+        Callback callback = new Callback(callbackNumber, runnable, CurrentAddress);
         _callbackHandler.AddCallback(callback);
         WriteCallback(callback.Index);
     }
@@ -84,7 +84,7 @@ public class MemoryAsmWriter : MemoryWriter {
         if (inMemoryAddressSwitcher.DefaultAddress is null) {
             throw new UnrecoverableException("Cannot write a FAR call to a null address.");
         }
-        inMemoryAddressSwitcher.PhysicalLocation = WriteFarCall(inMemoryAddressSwitcher.DefaultAddress.Value).ToPhysical();
+        inMemoryAddressSwitcher.PhysicalLocation = WriteFarCall(inMemoryAddressSwitcher.DefaultAddress.Value).Linear;
     }
 
     /// <summary>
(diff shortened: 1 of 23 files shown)


src/Spice86.Core/Emulator/InterruptHandlers/Common/MemoryWriter/MemoryAsmWriter.cs
```diff
@@ -95,7 +95,7 @@ public void WriteFarCallToSwitcherDefaultAddress(InMemoryAddressSwitcher inMemor
     public SegmentedAddress WriteFarCall(SegmentedAddress destination) {
         WriteUInt8(0x9A);
         // Make a copy of the address since it is going to be modified by our writes.
```
kevinferrare, line 97: Comment can be removed

src/Spice86.Shared/Emulator/Memory/SegmentedAddress.cs
```diff
@@ -34,17 +17,40 @@ public SegmentedAddress(ValueTuple<ushort, ushort> segmentOffset) : this(segment
...
+        return this < other ? -1 : this == other ? 0 : 1;
+    }
+
+    /// <summary>
+    /// Determines whether the current SegmentedAddress object is equivalent to another SegmentedAddress object.
+    /// </summary>
+    /// <param name="other">The object to compare with the current SegmentedAddress object.</param>
+    /// <returns>true if the objects are equivalent; otherwise, false.</returns>
+    public bool Equals(SegmentedAddress other) {
+        return Linear == other.Linear;
```
kevinferrare, line 53: This will break CfgCpu on some corner cases.
We store a map of SegmentedAddress -> Instruction and we want to have a different entry if the instruction is accessed via a different segment, because ultimately for some, this changes the way the instruction behaves.

src/Spice86.Shared/Emulator/Memory/SegmentedAddress.cs
```diff
@@ -34,17 +17,40 @@ public SegmentedAddress(ValueTuple<ushort, ushort> segmentOffset) : this(segment
     public SegmentedAddress(ushort segment, ushort offset) {
         Segment = segment;
         Offset = offset;
+        Linear = (uint)(Segment << 4) + Offset;
```
kevinferrare, line 20: Now SegmentedAddress fits in 64bits instead of 32.
This should still not have any negative impact on performance because the value can still be transported via one register, but how is the performance compared to before the change?

src/Spice86.Core/Emulator/InterruptHandlers/Common/MemoryWriter/MemoryAsmWriter.cs
JorisVanEijden: I thought so too at first. But it does explain why we should not move this to `return CurrentAddress`

kevinferrare: Indeed, my bad

kevinferrare (approved):

src/Spice86.Shared/Emulator/Memory/SegmentedAddress.cs
JorisVanEijden: No measruable difference, also with cfgcpu

PR #1109, opened 2025-03-22
fix: Don't crash with assert, throw instead

maximilien-noal:
### Description of Changes

Throws `UnrecoverablException `instead of using `Debug.Assert`

### Rationale behind Changes

Makes the UI not crash. We log the exception, and report it to the user instead.

### Suggested Testing Steps

Already tested with Wizardry 7 (which doesn't work ATM).

### DIFF
--- a/src/Spice86.Core/Emulator/Devices/Video/VideoMemory.cs
+++ b/src/Spice86.Core/Emulator/Devices/Video/VideoMemory.cs
@@ -2,6 +2,7 @@ namespace Spice86.Core.Emulator.Devices.Video;
 
 using Spice86.Core.Emulator.Devices.Video.Registers;
 using Spice86.Core.Emulator.Devices.Video.Registers.Graphics;
+using Spice86.Shared.Emulator.Errors;
 
 using System.Diagnostics;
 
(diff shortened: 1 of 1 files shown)


src/Spice86.Core/Emulator/Devices/Video/VideoMemory.cs
```diff
@@ -153,7 +154,14 @@ private void HandleWriteMode1(Register8 planeEnable, ReadOnlySpan<bool> writePla
 
     private void HandleWriteMode0(byte value, Register8 planeEnable,
         ReadOnlySpan<bool> writePlane, Register8 setResetEnable, Register8 setReset, uint offset) {
-        Debug.Assert(offset < 0x10000);
+        if(offset >= 0x10000) {
+            throw new UnrecoverableException(
+                $"""
+                The emulated program tried to write into a video memory address beyond 0x10000, which is forbidden: 0x{offset:X} !
```
kevinferrare, line 160: forbidden for mode 0?

maximilien-noal: I think so.

maximilien-noal: Info added to the message.

JorisVanEijden (approved):

PR #1133, opened 2025-04-05
hotfix: Valdiation does not block fields updates. Plus: Memory addressable limit validation (limit is A20Gate.EndOfHighMemoryArea), MemoryEdit/MemoryEditAddress validation, simplified all CanExecute methods

maximilien-noal:
### PR Summary

This pull request refactors property setters in all ViewModel classes dealing with input validation in order to avoid out of sync behavior between ViewModel bindings and textboxes values.

INotifyDataErrorInfo is supposed only to notify of errors, not block fields updates.

Plus, it contains the following validation enhancments, and related fixes:

- MemoryEditAddress and MemoryEditValue are validated now
- The memory edit command has a can execute
- IsMemoryRangeValid property is gone, replaced by a proper CanExecute
- All CanExecutes use _validationsErrors dictionary every time it makes sense to do so
- We validate the memory addresses to see if they are also addressable (ie. not beyond the end of the high memory area)
- FlowExceptionToTaskScheduler is set to True in the async relay command, so if an exception happens we can manage it instead of crashing. This crash should not happen.
- Memory class was fixed in ReadRam and WriteRam to avoid OutOfRangeException (tested in the memory view by going at and writing values at 0x10FFEA, which is close to 0x10FFEF, the end of the HMA).
- MemoryEdit feature wasn't very usable and is now fixed.

### DIFF
--- a/src/Spice86/ViewModels/MemoryViewModel.cs
+++ b/src/Spice86/ViewModels/MemoryViewModel.cs
@@ -285,7 +285,7 @@ public partial class MemoryViewModel : ViewModelWithErrorDialog {
     [NotifyCanExecuteChangedFor(nameof(GoToFoundOccurenceCommand))]
     private bool _isAddressOfFoundOccurrenceValid;
 
-    [RelayCommand(CanExecute = nameof(IsPaused), FlowExceptionsToTaskScheduler = false, IncludeCancelCommand = true)]
+    [RelayCommand(CanExecute = nameof(IsPaused), FlowExceptionsToTaskScheduler = true, IncludeCancelCommand = true)]
     private async Task SearchMemory(CancellationToken token) {
         if (string.IsNullOrWhiteSpace(MemorySearchValue) || token.IsCancellationRequested) {
             return;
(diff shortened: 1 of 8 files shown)


src/Spice86/ViewModels/MemoryViewModel.cs
```diff
@@ -115,16 +115,14 @@ private void SetSearchDataType(MemorySearchDataType searchDataType) {
...
-                IsMemoryRangeValid = false;
-                DataMemoryDocument = null;
+            ValidateAddressRange(_state, value, EndAddress,
+                nameof(StartAddress));
+            IsMemoryRangeValid = true;
+            if (SetProperty(ref _startAddress, value)) {
+                TryUpdateHeaderAndMemoryDocument();
             }
+            IsMemoryRangeValid = false;
+            DataMemoryDocument = null;
```
JorisVanEijden, line 125: Are you sure we want to do this every time StartAddress is set?

maximilien-noal: No, I'm re-thinking this bit of code. This should be a CanExecute.

maximilien-noal: Done.

maximilien-noal:
I tested this myself again and for me it's OK. It fixes several crashes and issues currently present on the master branch that I unfortunately didn't catch.

I really have to look into acceptance testing of the UI with specflow / reqnroll, but that's a lot of heavy work.

And there's already a lot of heavy work to do on Dune alone.

JorisVanEijden (approved):
I haven't tested it, but the changes look like they'll make things better, not worse.

PR #1134, opened 2025-04-05
adds small native DOS test examples for partially missing DOS APIs

LowLevelMahn:


tests/Spice86.Tests/Resources/NativeDosTests/build.bat
```diff
@@ -0,0 +1,18 @@
+rem grab your Open Watcom V2 copy from https://github.com/open-watcom/open-watcom-v2/releases
```
maximilien-noal, line 1: Can I compile this on Linux or macos ? Perhaps with dosbox staging ?

Another solution would be to include precompiled versions for unit tests in the future.

maximilien-noal (approved):
DOS and TSR examples written directly in assembly, that's *hardcore* !

This will be of great help and help move the DOS Process Managment PR to move forward.

tests/Spice86.Tests/Resources/NativeDosTests/build.bat
maximilien-noal: That's good.

Yes the pre-compiled DOS binaries would be nice, but I can do that ship thoses with the DOS process managment PR later.

LowLevelMahn: ok i will add the exes later

LowLevelMahn: just ask on discord if you need help building watcom on linux - is not that complex, some minutes to get your own fresh build

maximilien-noal: I'm on Windows so it will super easy to do with the .bat file you provided. :)

LowLevelMahn: yes just extract (not execute) the https://github.com/open-watcom/open-watcom-v2/releases/download/Current-build/open-watcom-2_0-c-win-x64.exe and fix the path in the batch

PR #1177, opened 2025-05-10
Int1A - SystemClock - Fix incorrectly assigned call numbers

ateska:
### Description of Changes

SystemClockInt1AHandler (INT 1A) has switched Get and Set calls.
This PR simply changes 0 <-> 1 to assign proper handlers to calls.

### Rationale behind Changes

https://stanislavs.org/helppc/int_1a.html

* `INT 1A` `AH=0`: Read system clock counter
* `INT 1A` `AH=1`:  Set system clock counter

### Suggested Testing Steps

I tested it on game Warlords.

Before the fix, it got stuck in the loading phase, printing "SET SYSTEM CLOCK COUNTER" in logs but in the debugger, I clearly see AH=0 for `INT 0x1A` syscall.

After the fix, I got into the main screen:

image


maximilien-noal (approved):
Hello,

Nice find of a nasty bug !

Thank you ! 😀

PR #1188, opened 2025-06-18
Feature/DOS IOCTL

maximilien-noal:
### Description of Changes

Solves DOS IOCTL operations compatibility, that a lot of games use. The DOS IOCTL is pretty much complete. Apart from specific classes that have yet to be created / fleshed out, like FloppyDiskDrive and a MSCDEX IOCTL driver (I'm *not* writing that one! ;-) )

Introduces a complete DOS Console Device, including ANSI support for colors based on characters escape sequences (see ANSI.SYS in the MS-DOS 5 manual).
It also supports the DOS INT21H function BufferedInput, that gets input from the keyboard. See class 'EmulationLoopRecalls'.

Programs can now print to a file or device in a generic way via INT21H, and it is logged.

There are more logs a little bit all over the place to help with making games work.

Introduces DOS Disk interrupt handlers INT25H and INT26H.

Solves bugs in DOS INT21H function ResizeBlock.

Adds some more small INT21H functions.

Makes SetZeroTerminatedString more performant.

Introduces BIOS INT13H Disk Interrupt Handler.

Adds GetShiftFlags in the INT16H handler.

Removes a bug in the Sound Blaster (it was registering DMA ports!)

Makes the INT10H Video Bios handler only log an Error in case VESA functions are called.
Some games call VESA functions, but resume if no one answers.
As I understand it, that's normal, a lot of cards were born before VESA 1.2 (aka VBE).

### Rationale behind Changes

Improves compatibility with DOS games. For example, Another World now runs.

### Suggested Testing Steps

Test your favorite game, and overrides. As on my end, they should still work.

I tested Duke Nukem, Duke Nukem II, Prince of Persia, Dune, Dune II, and Krondor without issues.


src/Spice86.Core/Emulator/InterruptHandlers/Bios/SystemBiosInt13Handler.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    private void FillDispatchTable() {
+        AddAction(0x0, () => ResetDiskSystem(true));
+    }
+
+    /// <summary>
+    /// Reset Disk System.
+    /// </summary>
+    /// <remarks>
+    /// This odes nothing and always clears CF.
```
kevinferrare: odes -> does?

src/Spice86.Core/Emulator/InterruptHandlers/Common/Callback/EmulationLoopRecalls.cs
```diff
@@ -0,0 +1,46 @@
...
+using Spice86.Core.Emulator.CPU;
+using Spice86.Core.Emulator.VM;
+using Spice86.Shared.Emulator.Memory;
+using Spice86.Core.Emulator.InterruptHandlers.Input.Keyboard;
+
+/// <summary>
+/// A class that can asynchronously wait for machine code to run. Called by C# code.
+/// </summary>
+/// <remarks>Primarly used by the emulator, for example when an interrupt handler needs to wait for another interrupt handler. This has to go through the emulation loop, otherwise the wait would block the emulation process.</remarks>
+public class EmulationLoopRecalls {
```
kevinferrare, line 13: I am not sure to understand the name "Recalls".
Since as of now this is "only" reading keyboard via int 16, can't that be renamed to something closer? Maybe you have other uses in mind for the future?
Not sure as well to understand what is asynchronous here.

src/Spice86.Core/Emulator/InterruptHandlers/Common/Callback/EmulationLoopRecalls.cs
```diff
@@ -0,0 +1,46 @@
...
+    private readonly InterruptVectorTable _interruptVectorTable;
+    private readonly SegmentedAddress _biosKeyboardCallback;
+    private readonly EmulationLoop _emulationLoop;
+    private readonly State _state;
+
+    public EmulationLoopRecalls(InterruptVectorTable ivt, State state, Stack stack, EmulationLoop emulationLoop) {
+        _nonLinearFlow = new(state, stack);
+        _state = state;
+        _emulationLoop = emulationLoop;
+        _interruptVectorTable = ivt;
```
kevinferrare, line 23: Looks like it could be a local variable

src/Spice86.Core/Emulator/InterruptHandlers/Common/Callback/NonLinearFlow.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Spice86.Core.Emulator.InterruptHandlers.Common.Callback;
+using Spice86.Core.Emulator.CPU;
+using Spice86.Shared.Emulator.Memory;
+
+internal class NonLinearFlow {
```
kevinferrare, line 5: Actually, this could be used by CPU and so on so maybe it shouldnt be internal?

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosDiskInt25Handler.cs
```diff
@@ -0,0 +1,39 @@
...
+        IFunctionHandlerProvider functionHandlerProvider, Stack stack, State state,
+        ILoggerService loggerService)
+        : base(memory, functionHandlerProvider, stack, state, loggerService) {
+        _dosDriveManager = dosDriveManager;
+    }
+
+    public override byte VectorNumber => 0x25;
+
+    public override void Run() {
+        if(State.AL >= DosDriveManager.MaxDriveCount || !_dosDriveManager.HasDriveAtIndex(State.AL)) {
```
kevinferrare: Even if some can be guessed from the code, how about using variables to make intent clearer?
AL => driveNumber
CX  => sectorToRead
DX => startingLogicalSector, 
new SegmentedAddress(State.DS, State.BX) => bufferForData

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosDiskInt25Handler.cs
```diff
@@ -0,0 +1,39 @@
...
+            State.AX = 0x8002;
+            State.CarryFlag = true;
+        } else {
+            if (State.CX == 1 && State.DX == 0) {
+                if (State.AL >= 2) {
+                    // write some BPB data into buffer for MicroProse installers
+                    Memory.UInt8[new SegmentedAddress(State.DS, (ushort)(State.BX + 0x1c))] = 0x3f; // hidden sectors
```
kevinferrare: There is also a UInt8 indexer that takes 2 parameters so no need to do a new?
I think this is for allowing installer to work well with copy protection right?

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -376,44 +380,126 @@ public void CloseFile(bool calledFromVm) {
...
+    /// Implements the INT21H function 0x3D, which reads standard input DOS Device and outputs it to the standard output DOS Device. <br/>
+    /// </summary>
+    /// <remarks>
+    /// TODO: Add check for Ctrl-C and Ctrl-Break in STDIN, and call INT23H if it happens.
+    /// </remarks>
+    public void BufferedInput() {
+        uint address = MemoryUtils.ToPhysicalAddress(State.DS, State.DX);
+        DosInputBuffer inputBufferHolder = new DosInputBuffer(Memory, address);
+        int readCount = 0;
+        if (!_dosFileManager.TryGetStandardInput(out CharacterDevice? standardInput) ||
```
kevinferrare, line 403: Not sure to understand well, those are stdin and stdout (or rather the device mapped there right?)

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -376,44 +380,126 @@ public void CloseFile(bool calledFromVm) {
...
+            if (readCount < 1) {
+                break; // No further input available, exit the loop.
+            }
+            byte c = inputBuffer[0];
+            if (c == (byte)AsciiControlCodes.LineFeed) {
+                continue;
+            }
+
+            if (c == (byte)AsciiControlCodes.Backspace) {
+                if (readCount != 0) { //Something to backspace.
```
kevinferrare, line 421: I think this cannot happen because you check count line 412

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
```diff
@@ -376,44 +380,126 @@ public void CloseFile(bool calledFromVm) {
...
 
+    /// <summary>
+    /// Implements the INT21H function 0x3D, which reads standard input DOS Device and outputs it to the standard output DOS Device. <br/>
+    /// </summary>
+    /// <remarks>
+    /// TODO: Add check for Ctrl-C and Ctrl-Break in STDIN, and call INT23H if it happens.
+    /// </remarks>
+    public void BufferedInput() {
+        uint address = MemoryUtils.ToPhysicalAddress(State.DS, State.DX);
+        DosInputBuffer inputBufferHolder = new DosInputBuffer(Memory, address);
```
kevinferrare: inputBufferHolder => inputBuffer?

src/Spice86.Core/Emulator/OperatingSystem/Devices/ConsoleDevice.cs
```diff
@@ -1,39 +1,608 @@
...
+                // What to do if buffer.Length == 1 and character is BackSpace ?
+                OutputWithNoAttributes(scanCode.Value);
+            }
+        }
+        _state.AX = oldAx; // Restore AX
+        return readCount;
+    }
+
+    /// <summary>
+    /// Tries to shortcut the emulation loop recall by reading memory first. <br/>
```
kevinferrare: Is it supposed to work this way? If not, I would just call int16, in case some crazy code overrides it.

src/Spice86.Core/Emulator/OperatingSystem/Dos.cs
```diff
@@ -197,11 +230,9 @@ public void AddDevice(IVirtualDevice device, ushort? segment = null, ushort? off
...
         if (device.Attributes.HasFlag(DeviceAttributes.Character)) {
-            var characterDevice = (CharacterDevice)device;
             _memory.LoadData(MemoryUtils.ToPhysicalAddress(device.Segment, index),
-                Encoding.ASCII.GetBytes( $"{characterDevice.Name,-8}"));
-        } else {
-            var blockDevice = (BlockDevice)device;
+                Encoding.ASCII.GetBytes( $"{device.Name,-8}"));
+        } else if(device is BlockDevice blockDevice) {
             _memory.UInt8[device.Segment, index] = blockDevice.UnitCount;
             index++;
```
kevinferrare, line 248: Not your code, but should really be done with a memory based structure, would make it much more obvious what is being done here.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -120,7 +156,33 @@ public DosFileOperationResult CreateFileUsingHandle(string fileName, ushort file
...
+        if (fileHandle == newHandle) {
+            return DosFileOperationResult.Error(ErrorCode.InvalidHandle);
+        }
+        if (!IsValidFileHandle(newHandle)) {
+            return DosFileOperationResult.Error(ErrorCode.InvalidHandle);
+        }
+        if (OpenFiles[newHandle] != null) {
+            return DosFileOperationResult.Error(ErrorCode.InvalidHandle);
+        }
+        if (GetOpenFile(fileHandle) is not VirtualFileBase file) {
```
kevinferrare, line 182: is not null? VirtualFileBase? is the return type of the method

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -120,7 +156,33 @@ public DosFileOperationResult CreateFileUsingHandle(string fileName, ushort file
...
+            return FileNotOpenedError(fileHandle);
+        }
+        if (newHandle < OpenFiles.Length && OpenFiles[newHandle] != null) {
```
kevinferrare, line 185: If newHandle >= OpenFiles.Length, the check a few lines above would have crashed isnt it?
(I am referring to line 175         if (OpenFiles[newHandle] != null) { )

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
```diff
@@ -185,11 +171,11 @@ private string GetFullDosPathIncludingRoot(string absoluteOrRelativeDosPath) {
             if (StartsWithDosDriveAndVolumeSeparator(dosPath)) {
                 length = 3;
             }
-            return (_driveMap[_currentDrive].MountedHostDirectory, dosPath[length..]);
+            return (_dosDriveManager.CurrentDrive.MountedHostDirectory, dosPath[length..]);
         } else if (StartsWithDosDriveAndVolumeSeparator(dosPath)) {
```
kevinferrare, line 175: Small logic error here maybe (not introduce by this PR):
IsPathRooted returns true if StartsWithDosDriveAndVolumeSeparator is true, so I guess this else if is never reached and we always resolve to current drive?

src/Spice86.Core/Emulator/InterruptHandlers/Common/Callback/EmulationLoopRecalls.cs
maximilien-noal: Yes, it's entirely sync code for us. Only async for the game, since this is an IO operation.

In dosbox staging the code is more generic and is called everywhere for different purposes. Except that it lets the caller setup the call via cpu registers. 

It made the code unclear I think. There will be other methods for those other purposes.

src/Spice86.Core/Emulator/InterruptHandlers/Common/Callback/NonLinearFlow.cs
maximilien-noal: With Internal it's public but just for the assembly. The CPU lives in tête same assembly so it can use it, unless I'm missing something ? 

Public will make it available to users too, Including future source compatibility / ABI breaking changes. I figured it was too 'low level' to be made public.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosDiskInt25Handler.cs
maximilien-noal: Yes this call should be clearer that way.

I think it's supposed to adress shortcomings in the referenced DOS or BIOS structure referenced in the comment. Without it the installers probably give an error.

This is also called by Lemmings 2: The Tribes, but it does not work yet and demands XMS

src/Spice86.Core/Emulator/Memory/MemoryRange.cs
```diff
@@ -39,7 +39,7 @@ public static MemoryRange FromSegment(ushort segmentStart, ushort length, string
     /// <returns>A new instance of the <see cref="MemoryRange"/> class.</returns>
     public static MemoryRange FromSegment(ushort segment, ushort startOffset, ushort length, string name) {
         uint start = MemoryUtils.ToPhysicalAddress(segment, startOffset);
-        uint end = MemoryUtils.ToPhysicalAddress(segment, (ushort)(startOffset + length));
```
maximilien-noal, line 42: Not so sure if this either introduce a bug or fixes a bug in the calculation of the end address.

What do you think @kevinferrare ?

src/Spice86.Core/Emulator/OperatingSystem/Devices/CharacterDevice.cs
```diff
@@ -1,22 +1,17 @@
...
-public class CharacterDevice : VirtualDeviceBase {
-    /// <summary>
-    /// The logging service.
-    /// </summary>
-    protected readonly ILoggerService Logger;
-
-    /// <summary>
-    /// 8-byte field with the name of the device.
-    /// </summary>
-    public string Name { get; }
```
maximilien-noal, line 19: This and Signature could use a Range attribute.

I think it ensures at runtime that limits are enforced. Rarely used it.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: Not very clear, I will replace it.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: OpenFiles has 255 entries. At startup they are all null.

Kinda weird but this code is good I think.

No dos game or overrides would open more than 0xff files.

src/Spice86.Core/Emulator/OperatingSystem/Devices/ConsoleDevice.cs
maximilien-noal: No this can be removed. It attempts an optimisation.

maximilien-noal: Code removed.

src/Spice86.Core/Emulator/InterruptHandlers/Bios/SystemBiosInt13Handler.cs
maximilien-noal: fixed.

src/Spice86.Core/Emulator/InterruptHandlers/Common/Callback/EmulationLoopRecalls.cs
maximilien-noal: fixed.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosInt21Handler.cs
maximilien-noal: done.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosDiskInt25Handler.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/InterruptHandlers/Dos/DosDiskInt25Handler.cs
maximilien-noal: Done.

src/Spice86.Core/Emulator/InterruptHandlers/Common/Callback/EmulationLoopRecalls.cs
maximilien-noal: Documentation fixed.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: The code has been clarified with comments. But it is the logical one.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
maximilien-noal: Yes, we use the same function to return InvalidHandle in our case.

Here, GetOpenFile(fileHandle) should always pass.

If it passes, we create a var named file and assign it to it.

We then used it a few lines below.

src/Spice86.Core/Emulator/OperatingSystem/DosPathResolver.cs
maximilien-noal: Yeah probably.

But this code makes me head hurt.

Is unit tested and always works, but I dare not touch it. :D

src/Spice86.Core/Emulator/OperatingSystem/Dos.cs
maximilien-noal: Done. Partially, at least.

PR #1195, opened 2025-07-01
feature: Implement the XADD instruction

rndtrash:
### Description of Changes

This PR adds support for the instructions `XADD r/m,r8` and `XADD r/m,r16/32` (opcodes `0F C0 /r` and `0F C1 /r`), which was introduced first for 486.

### Rationale behind Changes

I was trying to run a little game that was compiled with Borland Pascal 6.0, but the CPU class threw an exception somewhere in the code. There was a single instance of `xadd edx, edx` in the executable.

This patch has let me advance further, although not without problems, but that's outside the scope of this PR.

### Suggested Testing Steps

I am not sure how this code behaves with the `LOCK` prefix, but I don't know any real-life program that could be used to test that. However, here is the archive with that game:
[banana-quest.zip](https://github.com/user-attachments/files/20991411/banana-quest.zip)

### DIFF
--- a/src/Spice86.Core/Emulator/CPU/CPU.cs
+++ b/src/Spice86.Core/Emulator/CPU/CPU.cs
@@ -1,4 +1,4 @@
-﻿namespace Spice86.Core.Emulator.CPU;
+namespace Spice86.Core.Emulator.CPU;
 
 using Serilog.Events;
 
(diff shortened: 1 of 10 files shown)


kevinferrare (approved):

maximilien-noal:
Thanks ! Looks great.

Did you encounter more real mode games using 486 instructions ?

rndtrash:
Not yet, but I'll make an issue if I see one. It's probably going to be another one of those small Pascal games.

PR #1201, opened 2025-07-05
Fix/file access bug

maximilien-noal:
PR that fixes the DOS file spec being absent in the dos environment block.

Before (master):

image

After (fix/file_access_bug):

image

It also makes it a pointer in the PSP field at offset 0x2C. I put the actual env block arbitrarly at PspSegment + 1.

        byte[] environmentBlock = CreateEnvironmentBlock(file);
        // In the PSP, the Environment Block Segment field (defined at offset 0x2C) is a word, and is a pointer.
        int envBlockPointer = PspSegment + 1;
        SegmentedAddress envBlockSegmentAddress = new SegmentedAddress((ushort)envBlockPointer, 0);

        // Copy the environment block to memory in a separated segment.
        _memory.LoadData(MemoryUtils.ToPhysicalAddress(envBlockSegmentAddress.Segment,
            envBlockSegmentAddress.Offset), environmentBlock);

        // Point the PSP's environment segment to the environment block.
        psp.EnvironmentTableSegment = envBlockSegmentAddress.Segment;

This stops Monkey Island 2 and Prince of Persia 2 (and possibly other games) from asking for files that aren't there.

Those games do not run yet, but this has made quite a few games advance further in their code.

# Tests

Another World, Hocus Pocus, Wolf3D, Peter Pan, Shangai II, Krondor, Lost Eden, Duke Nukem, Duke Nukem II, Dune, Dune II, Stunts, Krondor, Prince of Persia, were tested without issues.

Hocus Pocus does not run on master branch.

### DIFF
--- a/src/Spice86.Core/Emulator/OperatingSystem/DosProcessManager.cs
+++ b/src/Spice86.Core/Emulator/OperatingSystem/DosProcessManager.cs
@@ -148,7 +148,7 @@ public class DosProcessManager : DosFileLoader {
         ms.WriteByte(0);
     
         // Get the DOS path for the program (not the host path)
-        string dosPath = GetDosProgramPath(programPath);
+        string dosPath = _fileManager.GetDosProgramPath(programPath);
     
         // Write the DOS path to the environment block
         byte[] programPathBytes = Encoding.ASCII.GetBytes(dosPath);
@@ -158,37 +158,7 @@ public class DosProcessManager : DosFileLoader {
         return ms.ToArray();
     }
 
-    /// <summary>
-    /// Gets the proper DOS path for a program.
-    /// </summary>
-    /// <param name="programPath">The program path as provided.</param>
-    /// <returns>A properly formatted DOS path.</returns>
-    private string GetDosProgramPath(string programPath) {
-        // Extract just the filename without path if it's a full path
-        string fileName = Path.GetFileName(programPath);
-        
-        // Create a DOS path using the current drive and directory
-        string currentDrive = _driveManager.CurrentDrive.DosVolume;
-        string currentDir = _driveManager.CurrentDrive.CurrentDosDirectory;
-        
-        // Ensure current directory has trailing backslash
-        if (!string.IsNullOrEmpty(currentDir) && !currentDir.EndsWith('\\')) {
-            currentDir += '\\';
-        }
-        
-        // Build the full DOS path
-        string dosPath = $"{currentDrive}\\{currentDir}{fileName}";
-        
-        // Replace slashes and standardize
-        dosPath = dosPath.Replace('/', '\\').ToUpperInvariant();
-        
-        // Clean up any double backslashes
-        while (dosPath.Contains("\\\\")) {
-            dosPath = dosPath.Replace("\\\\", "\\");
-        }
-        
-        return dosPath;
-    }
+    
 
     private byte[] LoadComFile(string file) {
         byte[] com = ReadFile(file);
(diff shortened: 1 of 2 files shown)


src/Spice86.Core/Emulator/LoadableFile/Bios/BiosLoader.cs
```diff
@@ -43,11 +43,11 @@ public BiosLoader(IMemory memory, State state, ILoggerService loggerService) : b
...
     /// <param name="arguments">Ignored for BIOS files.</param>
     /// <returns>The loaded BIOS file as a byte array.</returns>
-    public override byte[] LoadFile(string file, string? arguments) {
+    internal override (ushort CodeSegment, byte[] FileContent) LoadFile(string file, string? arguments) {
         byte[] bios = ReadFile(file);
         uint physicalStartAddress = MemoryUtils.ToPhysicalAddress(CodeSegment, 0);
         _memory.LoadData(physicalStartAddress, bios);
         SetEntryPoint(CodeSegment, CodeOffset);
-        return bios;
+        return (CodeSegment, bios);
```
kevinferrare: I would return the segmented address pointing to the entry point from which the caller can extract the segment

src/Spice86.Core/Emulator/OperatingSystem/Dos.cs
```diff
@@ -205,6 +208,12 @@ private void OpenDefaultFileHandles(VirtualFileBase[] fileDevices) {
         }
     }
 
+    internal void GeneratePspInMemory(Configuration configuration, ushort pspSegment, string? arguments) {
+        var pspGenerator = new PspGenerator(_memory, EnvironmentVariables, MemoryManager, FileManager);
+        string dosFileSpec = $"{Path.GetFileName(configuration.Exe)}".ToUpperInvariant();
```
kevinferrare: I would make the exe name a parameter so that in the future when we have subprograms we can use that instead of the conf

src/Spice86.Core/Emulator/LoadableFile/Bios/BiosLoader.cs
maximilien-noal: I changed strategy, we are using a DOS Process Manager class now. :)

src/Spice86.Core/Emulator/OperatingSystem/Dos.cs
maximilien-noal: This has been removed. :)

src/Spice86.Core/Emulator/OperatingSystem/DosProcessManager.cs
```diff
@@ -0,0 +1,270 @@
...
+    
+        return ms.ToArray();
+    }
+
+    /// <summary>
+    /// Gets the proper DOS path for a program.
+    /// </summary>
+    /// <param name="programPath">The program path as provided.</param>
+    /// <returns>A properly formatted DOS path.</returns>
+    private string GetDosProgramPath(string programPath) {
```
kevinferrare: Shouldn't this be part of DosFileManager?

kevinferrare (approved):

src/Spice86.Core/Emulator/OperatingSystem/DosProcessManager.cs
maximilien-noal: Done. But that's really only useful at startup time.

In the end, this should probably replaced once we generate AUTOEXEC.BAT and run it, but that's like, far in the future...

PR #1203, opened 2025-07-06
fix: performance measurement was degrading performance

maximilien-noal:
### Description of Changes

Fixes Lost Eden voices being unbearable to the ears.

### Rationale behind Changes

performance measurement was degrading performance

Details:

- Updated `EmulationLoop` to expose `CpuPerformanceMeasurer` and share it that way in dependency injection step with the PerformanceViewModel constructor.
- Split `IPerformanceMeasurer` into `IPerformanceMeasureReader` and `IPerformanceMeasureWriter`for better interface segragation.
- Updated `PerformanceViewModel` to use `IPerformanceMeasureReader` and accept a `cpuPerfReader` parameter.
- Removed `performanceMeasurer` parameter from `CfgCpuViewModel` constructor. It creates an instance internally, because this instance measures internal performance of the class.
- Improved code formatting in `PerformanceMeasurer` to respect 80 columns limit.

### Suggested Testing Steps

Already tested with Lost Eden and Dune and Krondor.

### DIFF
--- a/src/Spice86/ViewModels/PerformanceViewModel.cs
+++ b/src/Spice86/ViewModels/PerformanceViewModel.cs
@@ -7,12 +7,12 @@ using CommunityToolkit.Mvvm.ComponentModel;
 using Spice86.Core.Emulator.CPU;
 using Spice86.Core.Emulator.VM;
 using Spice86.Infrastructure;
-using Spice86.Shared.Diagnostics;
+using Spice86.Shared.Interfaces;
 
 using System;
 
 public partial class PerformanceViewModel : ViewModelBase {
-    private readonly PerformanceMeasurer _performanceMeasurer;
+    private readonly IPerformanceMeasureReader _cpuPerformanceReader;
     private readonly State _state;
 
     [ObservableProperty]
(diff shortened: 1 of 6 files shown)


src/Spice86/ViewModels/PerformanceViewModel.cs
```diff
@@ -7,37 +7,38 @@
...
+    public PerformanceViewModel(State state, IPauseHandler pauseHandler,
+        IUIDispatcher uiDispatcher, IPerformanceMeasureReader cpuPerfReader) {
+        _cpuPerformanceReader = cpuPerfReader;
         pauseHandler.Paused += () => uiDispatcher.Post(() => _isPaused = true);
         pauseHandler.Resumed += () => uiDispatcher.Post(() => _isPaused = false);
         _state = state;
         _isPaused = pauseHandler.IsPaused;
-        _performanceMeasurer = new PerformanceMeasurer();
-        DispatcherTimerStarter.StartNewDispatcherTimer(TimeSpan.FromSeconds(1.0 / 30.0), DispatcherPriority.Background, UpdatePerformanceInfo);
+        DispatcherTimerStarter.StartNewDispatcherTimer(TimeSpan.FromSeconds(0.4),
```
JorisVanEijden, line 30: Isn't this actually the biggest change in this pr? Going from 30 to 2.5 times per second?
I'm sure it's fine, just expected to see it in the description :)

JorisVanEijden:
I was not aware of a 80 column limit.
Personally I used to prefer 200. But now with an Ai window to the left, I use 120 or 140.

maximilien-noal:
> I was not aware of a 80 column limit. Personally I used to prefer 200. But now with an Ai window to the left, I use 120 or 140.

It's not a hard limit. 120/140 is fine too.

src/Spice86/ViewModels/PerformanceViewModel.cs
maximilien-noal: Yeah, after testing, just this line brings the same improvement.

PR #1204, opened 2025-07-06
1200 - Fix for writing the same character multiple times

### LINKED ISSUE #1200: Dune II text mode setup rendering bug
Image

JorisVanEijden:
### Description of Changes
When writing a batch of the same character multiple times, the character is now written seq4euantilly instead of all at the same position.

### Rationale behind Changes
Fixes #1200 and other programs that use the text functionality of writing a character _n_ times.

### Suggested Testing Steps
- Fire up Dune II's SETUP.EXE
- try any other text-mode program


maximilien-noal (approved):
Great fix !

It's so nice to be able to modify a game setup without dosbox.

PR #1205, opened 2025-07-06
Fix/Prince of Persia and Shangai II and MI1 read but didn't react to keyboard

maximilien-noal:
Those games were unplayable or could not save a game, etc...

Because the Console Device is using INT16H, but INT16H didn't wait for keyboard input and returned immediatly.

Shangai II, Dune, Dune II, Prince of Persia, Krondor, and Monkey Island 1 were tested without issue.


src/Spice86.Core/Emulator/InterruptHandlers/Common/Callback/EmulationLoopRecalls.cs
```diff
@@ -29,18 +34,18 @@ public EmulationLoopRecalls(InterruptVectorTable ivt, State state, Stack stack,
     /// <returns>Returns the scancode byte.</returns>
     public byte ReadBiosInt16HGetKeyStroke() {
```
JorisVanEijden: I think this should be called from Spice86.Core.Emulator.InterruptHandlers.Input.Keyboard.KeyboardInt16Handler.GetKeystroke().
That is the one that should watch the keyboard buffer to be filled by INT 9. (our current implementation is broken since it doesn't wait if the buffer is empty but returns 0x00)
Check out a real bios handling of INT 9 and 16: https://github.com/kaneton/appendix-bios/blob/master/src/keybd.asm 
INT 9: translate keyboard port data to scancode and asciicode for buffer
INT 16: query/read buffer
If we need to just run the CPU for interrupts while waiting for an entry in the keyboard buffer (int 16 00) we can write a small asm routine in our bios memory to do that.

maximilien-noal: Done, although not with ASM.

JorisVanEijden:
I looked at the PR a bit closer today and I don't think the ConsoleControl class is the correct way forward.
It cause some questionable coupling.
For instance, the DOS INT 21 functions should be device agnostic. Operating on stdin and stdout without knowledge of the console device.
Some examples:
- The `BufferedInput` method is descibed as "reads standard input DOS Device and outputs it to the standard output DOS Device"  now has "_consoleControl.Echo = true; // Enable echoing to the standard output."
- Calling the `ReadFromFileOrDevice` method now always enables echoing in the console device.
- The `DirectConsoleIo` method is modifying ConsoleControl regards of if the current StdOut is even the console device.

The way it is implemented also makes it impossible for me to see what problem it actually fixes.

maximilien-noal:
It's an unrelated change. I'll revert it.

maximilien-noal:
> The way it is implemented also makes it impossible for me to see what problem it actually fixes.

I reverted all the changes unrelated to the fix, and re-tested the games without issue.

PR #1213, opened 2025-07-19
fix out-of-bounds risk in GetDosString and simplify string handling

codengine:
An exception was thrown randomly while I tried to run "Another World". The end character was not found within 0..MaxDosStringLength. Strangely, I could not reproduce the issue afterwards but its better to fix it.

The string handling is also slightly improved.

I also thought about resizing the array if needed but would leave that open for discussion.

### Description of Changes
Prevent possible OutOfBounds exception

### Rationale behind Changes
Random exception while trying to run "Another World"

### Suggested Testing Steps
I just started "Another World" but somehow, the error disappeared.


maximilien-noal (approved):

maximilien-noal:
Thanks, I had this bug in mind but I could not correct it. Too much stuff to fix in DOS overall. Brain power was kaput. 😁

PR #1214, opened 2025-07-19
Fix/1208 video memory issue

JorisVanEijden:
### Description of Changes
We check if VGA memory access is within the currently mapped memory.
If it is not, we log writes and return 0x00 for reads.

### Rationale behind Changes
Dune2 writes debug information to 0xB0000+ which on real hardware would be visible on a secondary monochrome monitor.

### Suggested Testing Steps
Finish 1st mission of dune2.
Regression test your favorite game.

In the future we could implement this secondary screen.


maximilien-noal:
Thank you! :)

This fixes #1200 very nicely.

image

Also tested with Dune, no issues.

maximilien-noal (approved):

maximilien-noal:
Seems like all tests fail with this message:

`System.IndexOutOfRangeException : Index was outside the bounds of the array.`

maximilien-noal (changes requested):
Fixes the issue nicely, thank you ! :)

The only thing to fix is tests.

maximilien-noal (approved):

PR #1219, opened 2025-07-22
fix incorrect error code if FindFirstMatchingFile fails

codengine:
### Description of Changes
I debugged/traced how DosBox treats issues if a file can not be found in FindFirstMatchingFile. It turns out that the correct error code would is "NoMoreFiles". I also have a suspicion why the game does it. I assume there is a possibility to overwrite the games assets, which are in the data.-1- file, with files in the same directory. As we are returning the wrong error code, it would loop indefinitely.

In IDA, I can also see that it compares against 0x12 / dec 18:
grafik

Here is a proof that it works now.

grafik

### Rationale behind Changes
Improve compatibility by returning the correct error code.

### Suggested Testing Steps
Launch EnviroKids

### DIFF
--- a/src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
+++ b/src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
@@ -272,7 +272,7 @@ public class DosFileManager {
                 enumerationOptions);
 
             if (matchingPaths.Length == 0) {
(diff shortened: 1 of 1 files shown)


maximilien-noal:
I missed this one, great find!

maximilien-noal (approved):

PR #1226, opened 2025-07-25
Several compatibility fixes

codengine:
### Description of Changes
This PR will enhance the compatibility for multiple games, like "Das Telekommando" for example.

### Rationale behind Changes
Enhanced compatibility, fallback for undocumented/unsupported keyboard operations according to https://stanislavs.org/helppc/int_16.html

Mouse reset is required for "Batman Returns". 

The correct treatment of an unsupported keyboard feature is debatable, but in my opinion, it should be alright. I could not find a proper documentation in that case.

### DIFF
--- a/src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
+++ b/src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
@@ -54,10 +54,8 @@ public class KeyboardInt16Handler : InterruptHandler {
                 nameof(KeyboardInt16Handler), VectorNumber, operation);
         }
 
-        //If games that use those unsupported interrupts misbehave or crash, check if this state is the proper way to
-        //fix it as I couldn't find documentation about it
-        State.CarryFlag = true;
-        State.AX = 0;
+        //If games that use those unsupported interrupts misbehave or crash, check if certain flags/registers have to be set
+        //properly, e.g., AX = 0 and/or setting the carry flag accordingly.
     }
 
     /// <inheritdoc/>
--- a/src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
+++ b/src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
@@ -3,6 +3,7 @@
 using Serilog.Events;
 
 using Spice86.Core.Emulator.CPU;
+using Spice86.Core.Emulator.Errors;
 using Spice86.Core.Emulator.Function;
 using Spice86.Core.Emulator.InterruptHandlers.Common.Callback;
 using Spice86.Core.Emulator.InterruptHandlers.Common.MemoryWriter;
@@ -43,7 +44,17 @@ public class KeyboardInt16Handler : InterruptHandler {
         _biosKeyboardBuffer = biosKeyboardBuffer;
         AddAction(0x00, GetKeystroke);
         AddAction(0x01, () => GetKeystrokeStatus(true));
-        AddAction(0x02, () => GetShiftFlags());
+        AddAction(0x02, GetShiftFlags);
+        AddAction(0x03, () => Unsupported(0x03));
+        AddAction(0x04, () => Unsupported(0x04));
+        AddAction(0x05, () => Unsupported(0x05));
+        AddAction(0x10, () => Unsupported(0x10));
+        AddAction(0x11, () => Unsupported(0x11));
+        AddAction(0x12, () => Unsupported(0x12));
+    }
+
+    private void Unsupported(int operation) {
+        throw GenerateUnhandledOperationException(operation);
     }
 
     /// <inheritdoc/>
@@ -137,6 +148,22 @@ public class KeyboardInt16Handler : InterruptHandler {
     /// <inheritdoc/>
     public override void Run() {
         byte operation = State.AH;
-        Run(operation);
+
+        if (HasRunnable(operation)) {
+            Run(operation);
+        } else {
+            HandleUndocumentedInterrupt(operation);
+        }
+    }
+
+    private void HandleUndocumentedInterrupt(byte operation) {
+        if (LoggerService.IsEnabled(LogEventLevel.Warning)) {
+            LoggerService.Warning(
+                "{ClassName} INT {Int:X2} {operation}: Unhandled/undocumented keyboard interrupt called, will ignore",
+                nameof(KeyboardInt16Handler), VectorNumber, operation);
+        }
+
+        State.CarryFlag = true;
+        State.AX = 0;
     }
 }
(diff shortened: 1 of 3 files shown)


JorisVanEijden:
I prefer not to use that stanislavs site due to russion propaganda and misinformation. 

Looks like you're turning the unsupported operation handling upside down. Instead of an exception on all unhandled operations you are creating the exception for a few cases and ignoring (only logging) the rest.

The idea behind the exception is to catch behaviour we have not encountered before.
If a game uses a callback we don't implement yet, we implement it. If we don't know how (yet), the implementation logs that and returns something that allows the game to continue.
That way we know exactly which callbacks are actually being used by games we try to support, and where we still lack support.

maximilien-noal:
The reason why ems can log and ignore is because all ems functions are defined as optional in the LIM specs, and we implement all the important ones. But that's a unique situation.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
```diff
@@ -137,6 +148,22 @@ public void FlushKeyboardBuffer() {
...
-        Run(operation);
+
+        if (HasRunnable(operation)) {
+            Run(operation);
+        } else {
+            HandleUndocumentedInterrupt(operation);
+        }
+    }
+
+    private void HandleUndocumentedInterrupt(byte operation) {
```
maximilien-noal: This hides unimplemented interrupts with logs, I'd rather follow @kevinferrare design and remove this (and changes in the calling code) to let the exception crash the emulator on unimplemented behavior, or implement the operation.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseInt33Handler.cs
```diff
@@ -564,8 +564,18 @@ private void FillDispatchTable() {
         AddAction(0x1A, SetMouseSensitivity);
         AddAction(0x1B, GetMouseSensitivity);
         AddAction(0x1C, SetInterruptRate);
+        AddAction(0x21, Reset);
         AddAction(0x24, GetSoftwareVersionAndMouseType);
     }
+    
+    private void Reset() {
```
maximilien-noal, line 571: This hides unimplemented interrupts with mouse driver reset, I'd rather follow @kevinferrare design and remove this (and changes in the calling code) to let the exception crash the emulator on unimplemented behavior, or implement the operation.

maximilien-noal (changes requested):

codengine:
Thanks for reviewing it. The rationale behind the way how I treated the "undocumented" keyboard interrupts is because DosBox also just ignores them.

I modified the implementation to only ignore specific ones. The proper return value is still debatable, thats why I decided to leave it as a warning until it has been clarified fully.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseInt33Handler.cs
codengine: How so? I just implement the reset functionality.

maximilien-noal:
> Thanks for reviewing it. The rationale behind the way how I treated the "undocumented" keyboard interrupts is because DosBox also just ignores them.

Dosbox can do it because it implements 99 percent of what's important for games, programs, and Windows up to 3.11. 

For now copying this behavior can hide important errors, making them easier to miss. 


> I modified the implementation to only ignore specific ones. The proper return value is still debatable, thats why I decided to leave it as a warning until it has been clarified fully.

Is it possible to implement them instead ?

codengine:
> Is it possible to implement them instead ?

I'd love to if I was able to find a documentation how? The game (I can't recall which one unfortunately - tested too many recently) did just work fine and did not appear to have any side effects of this change.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/MouseInt33Handler.cs
maximilien-noal: You're right, sorry!

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
```diff
@@ -43,7 +43,21 @@ public KeyboardInt16Handler(IMemory memory, BiosDataArea biosDataArea,
...
+    private void Unsupported(int operation) {
+        if (LoggerService.IsEnabled(LogEventLevel.Warning)) {
+            LoggerService.Warning(
+                "{ClassName} INT {Int:X2} {operation}: Unhandled/undocumented keyboard interrupt called, will ignore",
+                nameof(KeyboardInt16Handler), VectorNumber, operation);
+        }
+
+        //If games that use those unsupported interrupts misbehave or crash, check if this state is the proper way to
+        //fix it as I couldn't find documentation about it
+        State.CarryFlag = true;
```
maximilien-noal: Unless documentation can be found on this call, I think it's better to remove those two lines.

Besides, I do not see DOSBox do this ?

I would remove those two lines that sets registers, to avoid causing a crash or potentially weird behavior.

maximilien-noal (commented):
Just two registers writes in Unsupported() that I could not find documentation for, and this will be good.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
maximilien-noal: Unless this is required by Batman Returns ?

codengine: I removed setting the flag/register. Will check if another game uses this interrupt and how it behaves.
Thats why I left the log on WARNING as the current treatment may not be correct.

maximilien-noal: Nothing at all in the RBIL or BIOS source codes about this call. Weird.

src/Spice86.Core/Emulator/OperatingSystem/DosFileManager.cs
```diff
@@ -150,6 +150,15 @@ public DosFileOperationResult CloseFile(ushort fileHandle) {
     /// <returns>A <see cref="DosFileOperationResult"/> with details about the result of the operation.</returns>
     /// <exception cref="UnrecoverableException"></exception>
     public DosFileOperationResult CreateFileUsingHandle(string fileName, ushort fileAttribute) {
+        CharacterDevice? device = _dosVirtualDevices.OfType<CharacterDevice>()
```
maximilien-noal, line 153: Nice fix, I forgot about this possibility when implementing DOS IOCTL.

maximilien-noal (approved):

PR #1234, opened 2025-08-01
Segmented address can change even if linear address is the same

kevinferrare:
### Description of Changes
Ensure segmented addresses are considered equals only when their segments and offsets are the same.
This makes cfgcpu create different instructions if they are accessed via different segmented addresses which is the correct behavior since address of an instruction can affect its execution result.
Added a test to check everything is fine.
Added some high level doc for cfgcpu

### Suggested Testing Steps
Test games run the same
Test performance is not affected
Already tested with dune / krondor / 3dbench2


maximilien-noal (approved):
LGTM ! :) Great work.

A20Gate and HMA, that's exactly what I struggle with currently. Nice to see more unit tests involving those memory features.

PR #1235, opened 2025-08-03
refactor: instantiate only one CPU, depending on the command line parameters

codengine:
### Description of Changes
This change will only instantiate one CPU, the other changes make it compatible with the remaining code base.
Code overrides will not work with CfgCpu as it uses a different architecture. @kevinferrare needs to implement the necessary functions for the not-implemented functions which are used by CSharpOverrideHelper (and perhaps some more) - that was far out of my scope.

### Rationale behind Changes
The code is now cleaner and it takes less memory.

### Suggested Testing Steps
Run Spice86 with either CPU enabled


src/Spice86/Views/Converters/BooleanInverter.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Globalization;
+
+/// <summary>
+///     Converts a boolean to a Visibility.
+/// </summary>
+/// <remarks>
+///     If the parameter is true, the converter inverts the boolean value.
+/// </remarks>
+public class BooleanInverter : IValueConverter {
```
maximilien-noal: No need for this converter, you can simply use ! in Avalonia AXAML.

Visiiblity is also a boolean in Avalonia.

src/Spice86.Core/Emulator/CPU/CfgCpu/CfgCpu.cs
```diff
@@ -108,6 +109,23 @@ private void HandleExternalInterrupt(ICfgNode toExecute) {
...
-        _executionContextManager.CurrentExecutionContext.FunctionHandler.Call(CallType.EXTERNAL_INTERRUPT, _state.IpSegmentedAddress, expectedReturn, null);
+        _executionContextManager.CurrentExecutionContext.FunctionHandler.Call(CallType.EXTERNAL_INTERRUPT,
+            State.IpSegmentedAddress, expectedReturn, null);
+    }
+
+    public Stack Stack => _instructionExecutionHelper.Stack;
+    public State State { get; }
+    public InterruptVectorTable InterruptVectorTable => _instructionExecutionHelper.InterruptVectorTable;
+
+    public void FarRet(ushort numberOfBytesToPop) {
```
maximilien-noal, line 120: If there is NotImplementedException added it means that the contract cannot be fuillfiled. That's problematic.

Instead, the ICPU interface could be reduced to the common surface between the two CPUs.

src/Spice86.Core/Emulator/CPU/CfgCpu/CfgCpu.cs
```diff
@@ -108,6 +109,23 @@ private void HandleExternalInterrupt(ICfgNode toExecute) {
         (SegmentedAddress target, SegmentedAddress expectedReturn) = _instructionExecutionHelper.DoInterrupt(externalInterruptVectorNumber.Value);
 
         _executionContextManager.SignalNewExecutionContext(target, expectedReturn);
-        _executionContextManager.CurrentExecutionContext.FunctionHandler.Call(CallType.EXTERNAL_INTERRUPT, _state.IpSegmentedAddress, expectedReturn, null);
+        _executionContextManager.CurrentExecutionContext.FunctionHandler.Call(CallType.EXTERNAL_INTERRUPT,
+            State.IpSegmentedAddress, expectedReturn, null);
+    }
+
+    public Stack Stack => _instructionExecutionHelper.Stack;
```
maximilien-noal, line 116: This exposes internal state of the class. This is best to be avoided.

src/Spice86.Core/Emulator/CPU/ICPU.cs
```diff
@@ -0,0 +1,12 @@
+namespace Spice86.Core.Emulator.CPU;
+
+using Spice86.Core.Emulator.Function;
+
+public interface ICPU : IInstructionExecutor, IFunctionHandlerProvider {
+    public Stack Stack { get; }
```
maximilien-noal, line 6: Don't exposte Stack, State, IVT to the outside world.

src/Spice86.Core/Emulator/CPU/ICPU.cs
```diff
@@ -0,0 +1,12 @@
...
+    public State State { get; }
+    InterruptVectorTable InterruptVectorTable { get; }
+    void FarRet(ushort numberOfBytesToPop);
```
maximilien-noal, line 9: If FarRet, InterruptRet, NearRet cannot be implemented, they should not be part of the contract.

src/Spice86.Core/Emulator/ReverseEngineer/CSharpOverrideHelper.cs
```diff
@@ -86,7 +86,7 @@ public class CSharpOverrideHelper {
     /// <summary>
     /// Gets the stack of the CPU.
     /// </summary>
-    public Stack Stack => Machine.Stack;
+    public Stack Stack => Machine.Cpu.Stack;
```
maximilien-noal: Pass Stack directly to the ctor. Don't use the Machine object.

See trainwrecks: https://wiki.c2.com/?TrainWreck

src/Spice86.Core/Emulator/VM/Machine.cs
```diff
@@ -54,23 +53,13 @@ public sealed class Machine : IDisposable {
     /// <summary>
     /// The emulated CPU.
     /// </summary>
-    public Cpu Cpu { get; }
```
maximilien-noal, line 57: Don't remove properties (State, Stack) inside the Machine object, they are there to be used by C# overrides of emulated program code.

src/Spice86/ViewModels/CfgCpuViewModel.cs
```diff
@@ -33,10 +32,10 @@ public partial class CfgCpuViewModel : ViewModelBase {
 
     [ObservableProperty] private bool _isCfgCpuEnabled;
 
-    public CfgCpuViewModel(Configuration configuration, ExecutionContextManager executionContextManager, IPauseHandler pauseHandler) {
+    public CfgCpuViewModel(ExecutionContextManager executionContextManager, IPauseHandler pauseHandler) {
         _executionContextManager = executionContextManager;
         _performanceMeasurer = new PerformanceMeasurer();
-        IsCfgCpuEnabled = configuration.CfgCpu;
+        IsCfgCpuEnabled = true;
```
maximilien-noal: I would not set this to enabled by default, but rather pass the Configuration object. It is the single source of truth to know if the CfgCpu is enabled.

src/Spice86/ViewModels/CpuViewModel.cs
```diff
@@ -25,13 +25,16 @@ public partial class CpuViewModel : ViewModelBase, IEmulatorObjectViewModel {
...
+    [ObservableProperty] private bool _isCfgCpuEnabled;
+
     public CpuViewModel(State state, IMemory memory, IPauseHandler pauseHandler, IUIDispatcher uiDispatcher) {
         _cpuState = state;
         _memory = memory;
         pauseHandler.Paused += () => uiDispatcher.Post(() => _isPaused = true);
         _isPaused = pauseHandler.IsPaused;
         pauseHandler.Resumed += () => uiDispatcher.Post(() => _isPaused = false);
         DispatcherTimerStarter.StartNewDispatcherTimer(TimeSpan.FromMilliseconds(400), DispatcherPriority.Background, UpdateValues);
+        IsCfgCpuEnabled = false;
```
maximilien-noal: I would not set this to enabled by default, but rather pass the Configuration object. It is the single source of truth to know if the CfgCpu is enabled.

src/Spice86/ViewModels/DebugWindowViewModel.cs
```diff
@@ -81,7 +79,7 @@ public DebugWindowViewModel(IMessenger messenger, IUIDispatcher uiDispatcher,
         MidiViewModel = midiViewModel;
         MemoryViewModels.Add(memoryViewModel);
         MemoryViewModels.Add(stackMemoryViewModel);
-        CfgCpuViewModel = cfgCpuViewModel;
+        SelectedTabIndex = cpuViewModel.IsCfgCpuEnabled ? 1 : 0;
```
maximilien-noal: Can the selectedTab be the dissassembly tab ?

tests/Spice86.Tests/CfgGraphDumper.cs
```diff
@@ -22,7 +20,8 @@ public List<string> ToAssemblyListing(Machine machine) {
     }
 
     private List<ICfgNode> DumpInOrder(Machine machine) {
-        Dictionary<SegmentedAddress, ISet<CfgInstruction>> executionContextEntryPoints = machine.CfgCpu.ExecutionContextManager.ExecutionContextEntryPoints;
+        Dictionary<SegmentedAddress, ISet<CfgInstruction>> executionContextEntryPoints =
+            ((Core.Emulator.CPU.CfgCpu.CfgCpu)machine.Cpu).ExecutionContextManager.ExecutionContextEntryPoints;
```
maximilien-noal, line 24: Can this cast be avoided ? Like by passing the execution context manager directly, like before

src/Spice86/Spice86DependencyInjection.cs
```diff
@@ -499,13 +498,10 @@ internal Spice86DependencyInjection(Configuration configuration, MainWindow? mai
 
             VideoCardViewModel videoCardViewModel = new(vgaRenderer, videoState);
 
-            CpuViewModel cpuViewModel = new(state, memory, pauseHandler, uiDispatcher);
-
+            ICpuViewModel cpuViewModel = cpuViewModelSupplier.Invoke(uiDispatcher);
```
maximilien-noal: Using an Action makes the code less understandable (at least for me).

Can the code be more direct ? Like use a local function, or an extension method ?

src/Spice86/Spice86DependencyInjection.cs
```diff
@@ -199,27 +196,30 @@ internal Spice86DependencyInjection(Configuration configuration, MainWindow? mai
...
-        if (loggerService.IsEnabled(LogEventLevel.Information)) {
-            loggerService.Information("CfgCpu created...");
-        }
+            cpuViewModelSupplier = uiDispatcher => new CpuViewModel(state, memory, pauseHandler, uiDispatcher);
+        } else {
+            var cfgCpu = new CfgCpu(memory, state, ioPortDispatcher, callbackHandler,
+                dualPic, emulatorBreakpointsManager, functionCatalogue, loggerService);
+            cpu = cfgCpu;
 
-        IInstructionExecutor instructionExecutor = configuration.CfgCpu ? cfgCpu : cpu;
```
maximilien-noal, line 218: Don't replace direct dependencies with indirect dependencies through the ICPU interface.

tests/Spice86.Tests/CfgGraphDumper.cs
```diff
@@ -22,7 +20,8 @@ public List<string> ToAssemblyListing(Machine machine) {
     }
 
     private List<ICfgNode> DumpInOrder(Machine machine) {
```
maximilien-noal, line 22: Don't reference the Machine anywhere, it's not to be used by the emulator.

It's just there as a convenience for C# emulated program code overrides, and is a legacy object.

tests/Spice86.Tests/MachineTest.cs
```diff
@@ -170,7 +169,8 @@ public void TestSegpr(bool enableCfgCpu)
         Machine machine = TestOneBin("segpr", enableCfgCpu);
         if (enableCfgCpu) {
             // Here, a division by 0 occurred causing a CPU fault. It is handled by an interrupt handler.
-            CurrentInstructions currentInstructions = machine.CfgCpu.CfgNodeFeeder.InstructionsFeeder.CurrentInstructions;
+            CurrentInstructions currentInstructions = ((Core.Emulator.CPU.CfgCpu.CfgCpu)machine.Cpu).CfgNodeFeeder
```
maximilien-noal, line 172: Can this cast be avoided, like before ?

tests/Spice86.Tests/MachineTest.cs
```diff
@@ -224,7 +224,8 @@ public void TestSelfModifyValue(bool enableCfgCpu)
         expected[0x03] = 0xff;
         Machine machine = TestOneBin("selfmodifyvalue", expected, enableCfgCpu);
         if (enableCfgCpu) {
-            CurrentInstructions currentInstructions = machine.CfgCpu.CfgNodeFeeder.InstructionsFeeder.CurrentInstructions;
+            CurrentInstructions currentInstructions = ((Core.Emulator.CPU.CfgCpu.CfgCpu)machine.Cpu).CfgNodeFeeder
```
maximilien-noal, line 227: Can this cast be avoided, like before ?

maximilien-noal (changes requested):
Hi,

Thank you for this PR. It should be a move in the right direction.

However, as it stands, I see several issues that need to be resolved:

- direct dependencies inside constructors replaced with indirected dependencies
- type casts that could be avoided
- interfaces that cannot be fullfilled
- usage of the Machine object (which is something I had to remove before in the entire codebase)
- not using Configuration to know wether the CfgCpu is enabled
- usage of Action.Invoke instead of a named method.
- IValueConverter for Visibility that could be replaced with a single '!' in XAML

src/Spice86.Core/Emulator/VM/Machine.cs
codengine: Its not removed but replaced with ICpu. Existing overrides will break because the CPU was exposed directly before.

src/Spice86/Views/Converters/BooleanInverter.cs
codengine: Thanks, I changed it.

src/Spice86.Core/Emulator/CPU/CfgCpu/CfgCpu.cs
codengine: When I do that, the CSharpOverrideHelper won't work anymore. The CSharpOverrideHelperTest can only work with CPUs that implement those three instructions.

I can make it so that the original "Cpu" has to be passed directly, but then the dependency injection in the CSharpOverrideHelperTest won't work anymore or I have to typecheck and cast the CPU created by the DI - something that you would most likely comment on.

What should I do?

src/Spice86.Core/Emulator/CPU/ICPU.cs
codengine: When I remove everything, then this MR is pointless.

src/Spice86.Core/Emulator/CPU/CfgCpu/CfgCpu.cs
codengine: Its just for the sake of compatibility with the other CPU.

src/Spice86.Core/Emulator/CPU/ICPU.cs
codengine: See my comment here https://github.com/OpenRakis/Spice86/pull/1235#discussion_r2250189312

src/Spice86/Spice86DependencyInjection.cs
codengine: I changed it to a switch.

src/Spice86/Spice86DependencyInjection.cs
codengine: Changed.

tests/Spice86.Tests/CfgGraphDumper.cs
codengine: It was like that before and out of scope for this MR.

tests/Spice86.Tests/MachineTest.cs
codengine: Without adding a type to Machine, unfortunately no. At least not with a lot of effort.

When you look at MachineTest, some tests are toggled off when enableCfgCpu is false.

tests/Spice86.Tests/MachineTest.cs
codengine: See https://github.com/OpenRakis/Spice86/pull/1235#discussion_r2250195681

src/Spice86.Core/Emulator/CPU/CfgCpu/CfgCpu.cs
kevinferrare: What I intended to do was to move this code from CPU to an external object and use that from cpu / cfgcpu. There is already such a class, by the way (used to re-enter emulation from C#, NonLinearFlow), to see if it's possible to use that instead.

codengine:
Closing this now. I think it makes more sense to remove the other cpu instead.

PR #1237, opened 2025-08-04
Improve headless mode

codengine:
The original implementation did not use a GUI, which made it so that applications behave differently in headless mode - some waiting for certain register values for example.

I noticed this difference in Space Job which throws an exception in GUI mode but just loops indefinitely in headless mode.

I also tried Avalonias Headless mode which also works fine but obviously requires more resources.

### Description of Changes
Introduce a headless GUI to simulate rendering at 60 Hz with a fake framebuffer.

### Rationale behind Changes
Improve headless mode to behave similarly to GUI mode.

### Suggested Testing Steps
Run some games in headless mode.


maximilien-noal (approved):
I wanted to improve headless mode and fix this difference in behavior. Never got the time to do it.

I reviewed the MR and it's very fine. Great job!

PR #1245, opened 2025-08-10
Make CFG CPU Dump execution flow

kevinferrare:
### Description of Changes
Make CFG CPU dump the execution flow in the same format as the regular CPU for compatibility with ghidra plugin.

This implies:
- Extracting the dumped json data structure to be useable by more than one implementation of data collector (and changing boilerplate code around that for regular CPU)
- Creating a new data dumper that can be used in place of the one used in CPU. It is only called when dumping data so no perf hit.
- Marking some cfginstruction classes to know whether they are call / ret / jump
- Handling cpu fault linking better

Node: We only dump the fields the fields used by the importer part of the plugin. Fields related to self modifying code in ghidra are not exported since the strategy we implemented in the ghidra plugin cant work well.

### Rationale behind Changes
This is part of the work needed to make cfgcpu mainstream

### Suggested Testing Steps
Tested with dune, 3dbench2 and krondor demo.

### DIFF
--- a/src/Spice86.Core/Emulator/CPU/CfgCpu/Feeder/CfgNodeFeeder.cs
+++ b/src/Spice86.Core/Emulator/CPU/CfgCpu/Feeder/CfgNodeFeeder.cs
@@ -8,8 +8,6 @@ using Spice86.Core.Emulator.CPU.CfgCpu.ParsedInstruction.SelfModifying;
 using Spice86.Core.Emulator.Memory;
 using Spice86.Core.Emulator.VM.Breakpoint;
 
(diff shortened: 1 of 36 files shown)


codengine:
I ran a test suite of 33 games against it and the results were similar to master.
So from that perspective, thumbs up from my side (I haven't looked at the changed yet though).

maximilien-noal:
The changes look great to me.

I didn't retest the gdb server manually in a long while, however. But the changes are unlikely to break it.

I should also test the cfgcpu dump, to see how it looks like.

Right now the heat is preventing me from starting up any machines...

PR #1247, opened 2025-08-11
fix: keyboard (oh no more lemmings) (NOT) (WIP)

### LINKED ISSUE #329: Feature request: implement  I/O 0x60 & 0x64, irq12, INT 74 and INT 15  (PS/2 mouse and keyboard)
Description of protocol between mouse and I/O ports 0x60 and 0x64:
    https://wiki.osdev.org/Mouse_Input

Assembly source code of ctmouse driver, including some very informative txt files:
    https://github.com/FDOS/mouse

Tell BIOS which routine to call when handling irq12 / INT 74
    INT 15 C2 07: https://mirror.cs.msu.ru/oldlinux.org/Linux.old/docs/interrupts/int-html/rb-1603.htm

IBM doc on keyboard/mouse controller:
    http://www.mcamafia.de/pdf/ibm_hitrc07.pdf

INT 15 handling BIOS code:
    
https://github.com/coreboot/seabios/blob/master/src/mouse.c

maximilien-noal:
Tries and utterly fails to fix #329 and also bugs such as the menu of Oh no! More Lemmings DOS console menu being next to impossible to use.

Maybe will fix Jill of the Jungle keyboard issues too.

Preview PR because everything is broken at the moment.

### DIFF
--- a/src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
+++ b/src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
@@ -7,6 +7,7 @@ using Spice86.Core.Emulator.Devices.ExternalInput;
 using Spice86.Core.Emulator.Devices.Input.Keyboard;
 using Spice86.Core.Emulator.Function;
 using Spice86.Core.Emulator.Memory;
+using Spice86.Shared.Emulator.Keyboard;
 using Spice86.Shared.Interfaces;
 
 /// <summary>
@@ -47,20 +48,32 @@ public class BiosKeyboardInt9Handler : InterruptHandler {
 
     /// <inheritdoc />
     public override void Run() {
-        byte? scanCode = _keyboard.KeyboardEvent.ScanCode;
-        byte ascii = _keyboard.KeyboardEvent.AsciiCode ?? 0;
-
-        if (scanCode is null) {
-            // No key pressed, nothing to enqueue
+        // Get the keyboard event directly rather than through port reads
+        // This ensures we're handling a single coherent event
+        KeyboardEventArgs keyboardEvent = _keyboard.GetNextKeyboardEvent();
+        
+        if (keyboardEvent.Equals(KeyboardEventArgs.None)) {
+            // No key event available
+            _dualPic.AcknowledgeInterrupt(1);
+            return;
+        }
+        
+        byte? scanCode = keyboardEvent.ScanCode;
+        byte ascii = keyboardEvent.AsciiCode ?? 0;
+        
+        if (!scanCode.HasValue) {
+            // No valid scan code
             _dualPic.AcknowledgeInterrupt(1);
             return;
         }
 
         if (LoggerService.IsEnabled(LogEventLevel.Verbose)) {
-            LoggerService.Verbose("{BiosInt9KeyReceived}", scanCode);
+            LoggerService.Verbose("{BiosInt9KeyReceived} ScanCode={ScanCode:X2} ASCII={ASCII:X2}", 
+                scanCode.Value, ascii);
         }
-
-        BiosKeyboardBuffer.EnqueueKeyCode((ushort)(scanCode << 8 | ascii));
+        
+        // Enqueue the key code into the BIOS keyboard buffer (as scancode << 8 | ascii)
+        BiosKeyboardBuffer.EnqueueKeyCode((ushort)(scanCode.Value << 8 | ascii));
         _dualPic.AcknowledgeInterrupt(1);
     }
 }
--- a/src/Spice86.Core/Emulator/VM/EmulationLoop.cs
+++ b/src/Spice86.Core/Emulator/VM/EmulationLoop.cs
@@ -2,6 +2,7 @@ namespace Spice86.Core.Emulator.VM;
 
 using Spice86.Core.Emulator.CPU;
 using Spice86.Core.Emulator.Devices.DirectMemoryAccess;
+using Spice86.Core.Emulator.Devices.Input.Keyboard;
 using Spice86.Core.Emulator.Devices.Timer;
 using Spice86.Core.Emulator.Errors;
 using Spice86.Core.Emulator.Function;
@@ -17,6 +18,7 @@ using System.Diagnostics;
 /// On Pause, triggers a GDB breakpoint.
 /// </summary>
 public class EmulationLoop {
+    private readonly IGui? _gui;
     private readonly ILoggerService _loggerService;
     private readonly IInstructionExecutor _cpu;
     private readonly FunctionHandler _functionHandler;
@@ -27,8 +29,7 @@ public class EmulationLoop {
     private readonly PerformanceMeasurer _performanceMeasurer;
     private readonly Stopwatch _stopwatch;
     private readonly DmaController _dmaController;
-
-    public IPerformanceMeasureReader CpuPerformanceMeasurer => _performanceMeasurer;
+    private readonly Keyboard _keyboard;
 
     /// <summary>
     /// Whether the emulation is paused.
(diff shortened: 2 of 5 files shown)


src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
```diff
@@ -6,35 +6,96 @@
...
         _dualPic = dualPic;
-        BiosKeyboardBuffer = biosKeyboardBuffer;
+    }
+
+    public override SegmentedAddress WriteAssemblyInRam(MemoryAsmWriter memoryAsmWriter) {
+        SegmentedAddress savedAddress = memoryAsmWriter.CurrentAddress;
+        memoryAsmWriter.CurrentAddress = CallbackLocation;
+
+        // push ax (save original AX)
+        memoryAsmWriter.WriteUInt8(0x50);
```
kevinferrare: Why cant this be c#?

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
```diff
@@ -6,35 +6,96 @@
...
+using Spice86.Shared.Emulator.Memory;
 using Spice86.Shared.Interfaces;
 
 /// <summary>
-/// Crude implementation of BIOS keyboard buffer handler (hardware interrupt 0x9, IRQ1)
+/// Implementation of BIOS keyboard buffer handler (hardware interrupt 0x9, IRQ1)
 /// </summary>
 public class BiosKeyboardInt9Handler : InterruptHandler {
-    private readonly Keyboard _keyboard;
+    private readonly AvaloniaKeyConverter _scanCodeConverter = new();
```
kevinferrare: Why is bios dealing with avalonia? shouldnt this be done at the device level?

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
maximilien-noal: Fewer dependencies. I don't have to depend on EmulationLoopRecall.

Plus it would be more code in C# and some parts like EOI stuff were easier this way.

TSR are also supposed to be able to react to this, this is easier if it's asm.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
maximilien-noal: This is temporary, the ps2 keyboard is supposed to have several scan code sets, and it's INT9H handler that is supposed to convert scan code into ASCII.

But this alone would bloat the PR by thousands of lines.

src/Spice86.Core/Emulator/Devices/Input/Keyboard/AvalionaKeyConverter.cs
```diff
@@ -20,9 +20,9 @@ internal class AvaloniaKeyScanCodeConverter {
     private static readonly FrozenDictionary<byte, byte> _scanCodeToAscii;
 
     /// <summary>
-    /// Initializes static members of the <see cref="AvaloniaKeyScanCodeConverter"/> class.
+    /// Initializes static members of the <see cref="AvaloniaKeyConverter"/> class.
     /// </summary>
-    static AvaloniaKeyScanCodeConverter() {
+    static AvaloniaKeyConverter() {
```
kevinferrare: There is a typo in the file name (avalionia)

src/Spice86.Core/Emulator/Devices/Input/Keyboard/PS2Keyboard.cs
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+        if (_loggerService.IsEnabled(LogEventLevel.Verbose)) {
+            _loggerService.Verbose("Processing keyboard event: {KeyboardEvent}, Scan Code: 0x{ScanCode:X2}", e, e.ScanCode);
+        }
+
+        AddScanCodeToBuffer(e.ScanCode.Value);
+    }
+
+    private void AddScanCodeToBuffer(byte scanCode) {
```
kevinferrare: Can't this be replaced by a fifo queue?

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
```diff
@@ -6,35 +6,51 @@
...
+    private readonly PS2Keyboard _keyboard;
+    private static readonly SegmentedAddress CallbackLocation = new(0xf000, 0xe987);
```
kevinferrare: Why does it needs to be precisely there?

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
```diff
@@ -47,20 +63,35 @@ public BiosKeyboardInt9Handler(IMemory memory,
 
     /// <inheritdoc />
     public override void Run() {
-        byte? scanCode = _keyboard.KeyboardEvent.ScanCode;
-        byte ascii = _keyboard.KeyboardEvent.AsciiCode ?? 0;
+        _keyboard.WriteByte(KeyboardPorts.Command, (byte)KeyboardCommand.DisablePortKbd);
```
kevinferrare, line 59: Shouldnt this logic to disable keyboard, read data and enable again be in keyboard class? Overrides might need it.
I wonder why doing so though, to ensure no other key gets added to the buffer? Wouldn't that solve the problem partially considering between time keyboard request irq1 and time this actually gets called, some time could already have passed? (Think about masked irq or other irq being processed)

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
```diff
@@ -47,20 +63,35 @@ public BiosKeyboardInt9Handler(IMemory memory,
...
+        byte scanCode = _keyboard.ReadByte(KeyboardPorts.Data);
+        _keyboard.WriteByte(KeyboardPorts.Command, (byte)KeyboardCommand.EnablePortKbd);
+
+        byte savedAl = State.AL;
```
kevinferrare, line 79: Ushort savedAX = State.AX since you use both al and ah?

src/Spice86.Core/Emulator/VM/EmulationLoop.cs
```diff
@@ -123,6 +121,7 @@ private void RunOnce() {
         _performanceMeasurer.UpdateValue(_cpuState.Cycles);
         _timer.Tick();
         _dmaController.PerformDmaTransfers();
+        _inputEventQueue.ProcessAllPendingInputEvents();
```
kevinferrare, line 131: Why cant this be done on ui thread that would put events in a queue in each device? Wouldn't this be less impactful than checking that every cycle?

src/Spice86/ViewModels/MainWindowViewModel.cs
```diff
@@ -137,12 +138,15 @@ private void RefreshMainTitleWithInstructionsPerMs() {
     internal void OnMainWindowClosing() => _isAppClosing = true;
 
     internal void OnKeyUp(KeyEventArgs e) {
+        if (IsPaused) {
```
kevinferrare: Can't this be checked in one common class that would be used to enquête events for the devices? UI would push events thrre and it would ignore them if paused

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/BiosKeyboardInt9Handler.cs
maximilien-noal: this was removed.

PR #1251, opened 2025-08-15
feat: CPU cycles limiting for real mode games

### LINKED ISSUE #1250: Bug: real mode speed sensitive games tend to crash
Somes games crash when CPU cycles are too high.

See: Lotus, Lotus 3, and this list: https://www.vogonswiki.com/index.php/List_of_CPU_speed_sensitive_games

maximilien-noal:
### Description of Changes

Fixes #1250

### Rationale behind Changes

Implement CPU cycles limiting for the rare speed sensitive real mode DOS game.

See this list: https://www.vogonswiki.com/index.php/List_of_CPU_speed_sensitive_games

The default value is 3000, because Dune CD is fine with that, and that is the default DOSBox setting.

By default, this feature isn't used - we don't limit performance in any way.

But using this also avoids:

* audio or video being underpowered / desyncing because the emulated CPU starves them of resources
* too much host CPU power used trying to go as fast as possible, while a limited < 10000 number of cycles per ms is enough
* UI being unresponsive because the emulated program is stuck in a loop, burning both the emulated and host CPU

### Suggested Testing Steps

Test a game that is supposed to be CPU speed sensitive with the `--Cycles` option.

### Tests results

Dune CD (full intro, game, cutscenes, videos, dialogues, gameplay, ornis) - with 3000 cycles is fine.
Duke Nukem II - fine with 3000 cycles
Dune II (3000 cycles works fine without making music slow down)
Monkey Island (kinda slow at 3000 cycles, fine with a little more, or without this feature being used)
Krondor (needs way more than 3000 cycles)
Prince of Persia (fine at 3000 cycles)

### DIFF
--- a/src/Spice86/Views/MainWindow.axaml
+++ b/src/Spice86/Views/MainWindow.axaml
@@ -30,7 +30,7 @@
 	<Design.DataContext>
 		<vm:MainWindowViewModel />
 	</Design.DataContext>
-	<Grid RowDefinitions="Auto,Auto,*,Auto">
+	<Grid RowDefinitions="Auto,*,Auto">
 		<Menu Name="Menu" Focusable="False" Grid.Row="0" IsVisible="{Binding !IsDialogVisible}">
 			<MenuItem Header="Debug">
 				<MenuItem HotKey="Ctrl+Alt+F2" ToolTip.Tip="Ctrl+Alt+F2" IsEnabled="{Binding IsEmulatorRunning}">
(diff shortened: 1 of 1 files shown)


src/Spice86.Core/CLI/Configuration.cs
```diff
@@ -9,6 +9,12 @@ namespace Spice86.Core.CLI;
 
 /// <summary> Configuration for spice86, that is what to run and how. Set on startup. </summary>
 public sealed class Configuration {
+    /// <summary>
+    /// CPU cycles per ms. Can be more precisely specified than instructions per second. Overrides instructions per second if used.
+    /// </summary>
+    [Option(nameof(Cycles), Default = null, Required = false, HelpText = "Precise control of the number of emulated CPU cycles per ms. For the rare speed-sensitive game. Default is 3000. Overrides instructions per second option if used.")]
```
kevinferrare: is default 3000 or null?

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
```diff
@@ -0,0 +1,107 @@
...
+    private const int MaxCyclesPerWindow = 20;
+    private const int CyclesUp = 1000;
+    private const int CyclesDown = 1000;
+    private const int MaxCyclesPerMs = 60000;
+    private const int MinCyclesPerMs = 100;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CpuCycleThrottler"/> class.
+    /// </summary>
+    /// <param name="targetCpuCyclesPerMs">The target CPU cycles per millisecond.</param>
```
kevinferrare: Should say that 0 means 3000?

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    /// <inheritdoc/>
+    public override void IncreaseCycles() {
+        TargetCpuCyclesPerMs = Math.Min(TargetCpuCyclesPerMs + CyclesUp,
+            MaxCyclesPerMs);
+    }
+
+    /// <inheritdoc/>
+    public override void DecreaseCycles() {
+        TargetCpuCyclesPerMs = Math.Max(TargetCpuCyclesPerMs - CyclesDown,
```
kevinferrare, line 104: What happens if this becomes 0?

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    /// <inheritdoc/>
+    internal override void RegulateCycles(State cpuState) {
+        if (!cpuState.IsRunning) {
+            return;
+        }
+
+        // If we still have cycle budget left, decrement and return
+        if (_cycleRemain > 0) {
+            _cycleRemain--;
```
kevinferrare: couldnt we avoid doing that and compare instead the cpuState.Cycles property with a _targetCyclesForNextMs var?

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
```diff
@@ -0,0 +1,107 @@
...
+            currentTicks = _stopwatch.ElapsedTicks;
+        }
+
+        // Calculate elapsed milliseconds
+        // (floating point for sub-millisecond precision)
+        double elapsedMs = (double)(currentTicks - _lastTicks)
+            / Stopwatch.Frequency * 1000;
+
+        // If we've advanced, update cycle budget
+        if (elapsedMs > 0) {
```
kevinferrare: What could be there cases where elapsed is 0?

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
```diff
@@ -0,0 +1,107 @@
...
+            _lastTicks = currentTicks;
+
+            // Calculate cycle budget with floating point to avoid losing
+            // sub-millisecond precision
+            _cycleRemain = (long)Math.Min(
```
kevinferrare: Why not just setting this to TargetCpuCyclesPerMs? I am probably missing a subtlety here.

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
```diff
@@ -0,0 +1,107 @@
...
+        if (_cycleRemain > 0) {
+            _cycleRemain--;
+            return;
+        }
+
+        // We've used all allocated cycles, need to calculate a new budget
+        long currentTicks = _stopwatch.ElapsedTicks;
+
+        // If time hasn't advanced significantly, make the emulation wait
+        if (currentTicks - _lastTicks < StopwatchTicksPerMillisecond) {
```
kevinferrare: So this is checked every ms? Do you think it could be made so that if precision is not too bad this is checked less and less frequently? (just a thought, but would that even make sense from cpu usage perspective?)

src/Spice86.Core/CLI/Configuration.cs
maximilien-noal: fixed documentation.

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
maximilien-noal: done.

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
maximilien-noal: It would make the CPU cycles vary more widely, making music sound bad. This is something I had trouble with for a long time with this branch.

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
maximilien-noal: IO delays in Debug mode, verbose logs, host CPU being underpowered or used by another process...

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
maximilien-noal: TargetCpuCyclesPerMs is defined by the user and what we constantly aim for.

_cyclesRemain is our budget of cycles

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
maximilien-noal: I don't think so ?

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
maximilien-noal: Actually it doesn't happen, and doesn't make sense. Removed.

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
maximilien-noal: done.

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleThrottler.cs
maximilien-noal: done.

src/Spice86.Core/Emulator/VM/CpuSpeedLimit/CpuCycleLimiter.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        long wallClockTicks = _stopwatch.ElapsedTicks;
+
+        // (floating point for sub-millisecond precision)
+        double elapsedMs = (double)(wallClockTicks - _lastTicks) / Stopwatch.Frequency * 1000;
+
+        _lastTicks = wallClockTicks;
+
+        // Calculate how many cycles we should allow before the next pause
+        long cyclesToAdd = (long)(TargetCpuCyclesPerMs * elapsedMs);
```
kevinferrare: Not sure I understand the logic here.
If elapsedMs was actually 2ms (because PC was slower for example to execute 3000 instructions), it means we will allow 6000? Is my understanding correct? If so why?

maximilien-noal: File IO or sharing host resources with other processus can wreck performance. It makes the cpu performance more stable over time.

I tried without it, it makes the performance vary widely.

maximilien-noal: Fixed.

PR #1257, opened 2025-08-17
Determine whether a DOS executable is a COM or EXE file based on file content and not extension

kevinferrare:
### Description of Changes
Some programs ship as COM files but they are actually EXE files (MZ header).
Let's not care about extension to differentiate the formats.

### Suggested Testing Steps
Working with stunts.com which is an EXE file although it has a COM extension
Working with dune / krondor


maximilien-noal:
Very nice fix!

PR #1289, opened 2025-08-21
Ensure memory breakpoints are triggered when a slice of memory is req…

kevinferrare:
…uested


src/Spice86.Core/Emulator/OperatingSystem/DosStringDecoder.cs
```diff
@@ -57,7 +58,7 @@ public string ConvertDosChars(Span<byte> characterBytes) {
     public string GetDosString(ushort segment, ushort offset, char end) {
```
maximilien-noal: It seems that there's no need to change this class after all.

I copied the entire class from the master branch and it still compiles and works.

Plus the new code allocates more memory on the heap (new List, .ToArray )

maximilien-noal (changes requested):
I tested a lot of games quickly this morning between 6 and 8 AM.

I got an InvalidOpCodeException with Dune 2 v1.0 (SB OPL music acive - SoundBlaster Pro selected in game SETUP.EXE, XMS enabled in game SETUP + CFGPU enabled), but could not get it repeatedly.

However, here is the logfile which does mention it.

[log-20250822.txt](https://github.com/user-attachments/files/21930845/log-20250822.txt)

Cadaver froze with the CFGPU, but it's the same on master.

src/Spice86.Core/Emulator/OperatingSystem/DosStringDecoder.cs
kevinferrare: It works if you duplicate method ConvertDosChars to have a version that accepts bytes and a version that accepts chars. I changed to this. I wouldn't worry about allocations though, this is not on the critical path

maximilien-noal:
More testing shows that I can't trigger any crash anymore with Dune 2 1.0 + CFGCPU + XMS.

Great job !

PR #1300, opened 2025-09-02
feat: internal breakpoint serialization

### LINKED ISSUE #1167: Debugger / Emulator: session support (save breakpoints on disk, essentially)
They will be stored in a folder besides the spice86 executable. Along with all the files we currently save (ie. logs, dumps, ...)

The session identifier is the DOS executable SHA256 signature.

The model to save on disk will probably have versionning...

A lot of stuff to be determined.

One thing is clear: Breakpoints will be saved on exit, and restored on startup of the same DOS executable.

To be tested with Dune and Cryogenic, and Krondor.

maximilien-noal:
### Description of Changes

Makes internal debugger breakpoints serializable, and saves/loads them automatically.

Reuses the existing infrastructure (ie. EmulatorStateSerializer class, and the EmulatorBreakpointsManager).

Closes #1167

### Rationale behind Changes

Stop redoing breakpoints each time you restart Spice86. Who's got the time for that ?! :)

We simply use the RecordedDataDirectory, and a file name set in stone: 'Breakpoints_{0}.json"

Where the 0 is the SHA256 hash of the executable being debugged.

We also store the SHA256 hash in the file, to ensure that our breakpoints are for the currently emulated program.

### Suggested Testing Steps

Already tested. PR works.
Including range breakpoints this time.

Tests show the same picture and behavior before and after restart:

Before serialization :

image

After deserialization:

image

Comments (auto-generated, not editable) are not serialized.

But eh, it works!

### DIFF
--- a/src/Spice86.Core/Emulator/VM/Breakpoint/EmulatorBreakpointsManager.cs
+++ b/src/Spice86.Core/Emulator/VM/Breakpoint/EmulatorBreakpointsManager.cs
@@ -237,4 +237,8 @@ public sealed class EmulatorBreakpointsManager : ISerializableBreakpointsSource 
             _ => throw new InvalidOperationException("Cannot deserialize unrecognized BreakpointType"),
         };
     }
+
+    public void RemoveUserBreakpoint(BreakPoint breakPoint) {
+        ToggleBreakPoint(breakPoint, false);
+    }
 }
--- a/src/Spice86/ViewModels/BreakpointsViewModel.cs
+++ b/src/Spice86/ViewModels/BreakpointsViewModel.cs
@@ -488,7 +488,7 @@ public partial class BreakpointsViewModel : ViewModelBase {
         if (breakpoint is null) {
             return;
         }
-        breakpoint.Disable();
+        breakpoint.Delete();
         Breakpoints.Remove(breakpoint);
         BreakpointDeleted?.Invoke(breakpoint);
     }
--- a/src/Spice86/ViewModels/BreakpointRangeViewModel.cs
+++ b/src/Spice86/ViewModels/BreakpointRangeViewModel.cs
@@ -6,6 +6,8 @@ using Spice86.Shared.Emulator.VM.Breakpoint;
 using System;
 
 public partial class BreakpointRangeViewModel : BreakpointViewModel {
+    private readonly List<BreakPoint> _breakpoints = new List<BreakPoint>();
+
     public BreakpointRangeViewModel(BreakpointsViewModel breakpointsViewModel,
         EmulatorBreakpointsManager emulatorBreakpointsManager, long trigger, long endTrigger,
         BreakPointType type, bool isRemovedOnTrigger, Action onReached,
@@ -13,6 +15,12 @@ public partial class BreakpointRangeViewModel : BreakpointViewModel {
         : base(breakpointsViewModel, emulatorBreakpointsManager, trigger, type,
             isRemovedOnTrigger, onReached, comment) {
         EndTrigger = endTrigger;
+        for (long i = Address; i <= EndTrigger; i++) {
+            AddressBreakPoint breakpoint = CreateBreakpointWithAddress(i);
+            breakpoint.IsEnabled = true;
+            _emulatorBreakpointsManager.ToggleBreakPoint(breakpoint, on: breakpoint.IsEnabled);
+            _breakpoints.Add(breakpoint);
+        }
     }
 
     public long EndTrigger { get; }
@@ -21,8 +29,8 @@ public partial class BreakpointRangeViewModel : BreakpointViewModel {
         if(!IsEnabled) {
             return;
         }
-        for(long i = Address; i <= EndTrigger; i++) {
-            DisableInternal(CreateBreakpointWithAddress(i));
+        foreach (BreakPoint breakpoint in _breakpoints) {
+            breakpoint.IsEnabled = false;
         }
         OnPropertyChanged(nameof(IsEnabled));
     }
@@ -31,8 +39,8 @@ public partial class BreakpointRangeViewModel : BreakpointViewModel {
         if(IsEnabled) {
             return;
         }
-        for (long i = Address; i <= EndTrigger; i++) {
-            EnableInternal(CreateBreakpointWithAddress(i));
+        foreach (BreakPoint breakpoint in _breakpoints) {
+            breakpoint.IsEnabled = true;
         }
         OnPropertyChanged(nameof(IsEnabled));
     }
--- a/src/Spice86/ViewModels/BreakpointViewModel.cs
+++ b/src/Spice86/ViewModels/BreakpointViewModel.cs
@@ -8,9 +8,9 @@ using Spice86.Shared.Emulator.Memory;
 using Spice86.Shared.Emulator.VM.Breakpoint;
 
 public partial class BreakpointViewModel : ViewModelBase {
+    private readonly BreakPoint _breakpoint;
     protected readonly Action _onReached;
-    private readonly EmulatorBreakpointsManager _emulatorBreakpointsManager;
-    protected readonly BreakPoint _breakPoint;
+    protected readonly EmulatorBreakpointsManager _emulatorBreakpointsManager;
 
     public BreakpointViewModel(
         BreakpointsViewModel breakpointsViewModel,
@@ -34,8 +34,10 @@ public partial class BreakpointViewModel : ViewModelBase {
         }
         Comment = comment;
         Parameter = $"0x{trigger:X2}";
-        _breakPoint = GetOrCreateBreakpoint();
-        _emulatorBreakpointsManager.ToggleBreakPoint(_breakPoint, _breakPoint.IsEnabled);
+        _breakpoint = GetOrCreateBreakpoint();
+        _breakpoint.IsEnabled = true;
+        _isEnabled = true;
+        _emulatorBreakpointsManager.ToggleBreakPoint(_breakpoint, _breakpoint.IsEnabled);
     }
 
     [ObservableProperty]
@@ -93,12 +95,12 @@ public partial class BreakpointViewModel : ViewModelBase {
         if (IsEnabled) {
             return;
         }
-        EnableInternal(_breakPoint);
+        EnableInternal(_breakpoint);
         OnPropertyChanged(nameof(IsEnabled));
     }
 
     protected void EnableInternal(BreakPoint breakpoint) {
-        breakpoint.IsEnabled = false;
+        breakpoint.IsEnabled = true;
         _isEnabled = true;
     }
 
@@ -112,7 +114,11 @@ public partial class BreakpointViewModel : ViewModelBase {
         if (!IsEnabled) {
             return;
         }
-        DisableInternal(_breakPoint);
+        DisableInternal(_breakpoint);
         OnPropertyChanged(nameof(IsEnabled));
+    }
+
+    internal void Delete() {
+        _emulatorBreakpointsManager.RemoveUserBreakpoint(_breakpoint);
     }
 }


src/Spice86.Core/Emulator/CPU/CfgCpu/Feeder/CurrentInstructions.cs
```diff
@@ -5,6 +5,7 @@ namespace Spice86.Core.Emulator.CPU.CfgCpu.Feeder;
 using Spice86.Core.Emulator.VM;
 using Spice86.Core.Emulator.VM.Breakpoint;
 using Spice86.Shared.Emulator.Memory;
+using Spice86.Shared.Emulator.VM.Breakpoint;
```
kevinferrare, line 8: Is this import needed?

src/Spice86.Core/Emulator/Function/Dump/EmulatorStateSerializer.cs
```diff
@@ -13,33 +13,41 @@ public class EmulatorStateSerializer {
...
     /// </summary>
-    public EmulatorStateSerializer(MemoryDataExporter memoryDataExporter, State state,
-        IExecutionDumpFactory executionDumpFactory, FunctionCatalogue functionCatalogue, ILoggerService loggerService) {
+    public EmulatorStateSerializer(
+        MemoryDataExporter memoryDataExporter, 
+        State state,
+        IExecutionDumpFactory executionDumpFactory, 
+        FunctionCatalogue functionCatalogue, 
+        ILoggerService loggerService,
+        EmulatorBreakpointsSerializer emulatorBreakpointsSerializer) {
```
kevinferrare: Logically the argument could be next to executionDumpFactory

src/Spice86.Core/Emulator/Gdb/GdbServer.cs
```diff
@@ -145,7 +145,7 @@ private void RunServer() {
     /// </summary>
     public void StartServer() {
         _gdbServerThread = new(RunServer) {
-            Name = "GdbServer"
+            Name = "GdbServer",
```
kevinferrare: trailing comma is unneeded?

src/Spice86.Core/Emulator/CPU/CfgCpu/Feeder/CurrentInstructions.cs
maximilien-noal: Yes. It's because the BreakPointType enum was moved in the Spice86.Shared assembly.

src/Spice86.Core/Emulator/Gdb/GdbServer.cs
maximilien-noal: fixed.

src/Spice86.Core/Emulator/Function/Dump/EmulatorStateSerializer.cs
maximilien-noal: done.

src/Spice86.Core/Emulator/CPU/CfgCpu/Feeder/CurrentInstructions.cs
maximilien-noal: If I remove this import, I get build errors.

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/EmulatorBreakpointsSerializer.cs
```diff
@@ -0,0 +1,115 @@
...
+    private ISerializableBreakpointsHolder? _serializableBreakpointsHolder = null;
+    private readonly string _programHash;
+
+    public EmulatorBreakpointsSerializer(Configuration configuration, ILoggerService loggerService) {
+        _configuration = configuration;
+        _loggerService = loggerService;
+        _programHash = GetProgramHash(configuration);
+    }
+
+    private string GetProgramHash(Configuration configuration) {
```
kevinferrare: The configuration value is just there for one purpose which is to restrict execution in case hash don't match. This is mainly used for overrides.

I suggest you make spice86 compute the exe hash all the time (in a separate class) and use that to:
- Check if param matches actual exe hash (current use case)
- Generate the folder name (this use case)

Generation of folder name shouldn't be there though, it should be done in case there is no configuration given to specify the dump directory (RecordedDataDirectory parameter)

Not handling all that here will simplify this code and will on top of that make dumping / loading execution data better. Maybe with this, this class could become very simple and become methods of RecordedDataReader / RecordedDataWriter?

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/EmulatorBreakpointsSerializer.cs
```diff
@@ -0,0 +1,115 @@
...
+using System;
+using System.Text.Json;
+using System.Text;
+using System.IO;
+using System.Security.Cryptography;
+
+/// <summary>
+/// Serializes (on startup) and deserialize (on app shutdown) internal debugger breakpoints from disk.
+/// </summary>
+public class EmulatorBreakpointsSerializer {
```
kevinferrare: I would rename it to UserBreakpointsSerializer or something like that, this would make it clear that we dont serialize internal breakpoints that the emulator creates

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
```diff
@@ -0,0 +1,50 @@
...
+    public bool IsEnabled { get; init; }
+
+    /// <summary>
+    /// Gets the condition expression for the breakpoint.
+    /// </summary>
+    /// <remarks>
+    /// Reserved for future use. This would allow expressions like "eax == 0x1234" 
+    /// to be evaluated when the breakpoint address is hit.
+    /// </remarks>
+    public string? Condition { get; init; }
```
kevinferrare: Until we support that, I would not add it here

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// <summary>
+    /// Gets a value indicating whether the breakpoint should be removed after being triggered.
+    /// </summary>
+    public bool IsRemovedOnTrigger { get; init; }
+
+    /// <summary>
+    /// Gets the comment for the breakpoint.
+    /// </summary>
+    public string Comment { get; init; } = string.Empty;
```
kevinferrare: This doesnt seem to be user fillable. Only use case I could find was the step over breakpoint.

By the way, do we serialize step over breakpoints with this code? I am not sure we would because they are supposed to be ephemeral usually, but what if step over is never reached and we save? Do we get them in the dump?

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// <summary>
+    /// Gets the type of the breakpoint.
+    /// </summary>
+    public BreakPointType Type { get; init; }
+
+    /// <summary>
+    /// Gets a value indicating whether the breakpoint should be removed after being triggered.
+    /// </summary>
+    public bool IsRemovedOnTrigger { get; init; }
```
kevinferrare: User breakpoints are not usually removed on trigger right?

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
```diff
@@ -0,0 +1,50 @@
...
+/// Base class for serializable breakpoint data.
+/// </summary>
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
+[JsonDerivedType(typeof(SerializedBreakpoint), typeDiscriminator: "breakpoint")]
+[JsonDerivedType(typeof(SerializedBreakpointRange), typeDiscriminator: "range")]
+public record SerializedBreakpoint {
+    /// <summary>
+    /// Gets the format version.
+    /// </summary>
+    public int Version { get; init; } = 1;
```
kevinferrare: What happens if version of the file does not match what we expect?
- If file grammar is different, we will crash before we can even read the content
- If grammar is the same, do you expect data to have different meaning in the future?

src/Spice86/ViewModels/BreakpointsViewModel.cs
```diff
@@ -470,4 +473,115 @@ private void DeleteBreakpoint(BreakpointViewModel? breakpoint) {
...
+    /// <returns>A SerializedBreakpoints object containing all the internal debugger breakpoints.</returns>
+    public SerializedBreakpoints CreateSerializableBreakpoints() {
+        var serializedBreakpoints = new SerializedBreakpoints {
+            Breakpoints = new List<SerializedBreakpoint>()
+        };
+
+        foreach (BreakpointViewModel breakpoint in Breakpoints) {
+            SerializedBreakpoint serializedBreakpoint;
+
+            if (breakpoint is BreakpointRangeViewModel rangeBreakpoint) {
```
kevinferrare, line 489: Shouldnt this logic be done in BreakpointViewModel?

maximilien-noal:
Back in preview as a preventive measure - range breakpoint serialization and/or deserialization is bugged.

src/Spice86/ViewModels/BreakpointsViewModel.cs
```diff
@@ -470,4 +473,115 @@ private void DeleteBreakpoint(BreakpointViewModel? breakpoint) {
...
+        foreach (SerializedBreakpoint breakpointData in breakpointsData.Breakpoints) {
+            if (breakpointData is SerializedBreakpointRange rangeData) {
+                RestoreRangeBreakpoint(rangeData);
+            } else {
+                RestoreBreakpoint(breakpointData);
+            }
+        }
+    }
+
+    private void RestoreBreakpoint(SerializedBreakpoint breakpointData) {
```
kevinferrare: I am not sure there is a good way because I dont know this part of the code well, but do you think there is a way to unify breakpoint creation from UI (CreateMemoryBreakpointAtAddress method for instance) and from SerializedBreakpoint object?

src/Spice86.Core/Emulator/Function/Dump/EmulatorStateSerializer.cs
```diff
@@ -13,33 +12,41 @@ public class EmulatorStateSerializer {
...
                 _state,
                 _executionDumpFactory,
                 _memoryDataExporter,
                 _functionCatalogue,
-                path, _loggerService)
+                path, 
+                _loggerService)
             .DumpAll();
+            
+        _emulatorBreakpointsSerializer.SaveBreakpoints();
```
codengine: It's unexpected that this method also serializes the breakpoints, especially because the provided path is not considered during serialization.

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/EmulatorBreakpointsSerializer.cs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public void SaveBreakpoints() {
+        string fileName = string.Format(BreakpointsFileNameFormat, _programHash);
+        SerializeBreakpoints(Path.Combine(_configuration.RecordedDataDirectory, fileName));
+    }
+
+    private void SerializeBreakpoints(string filePath) {
+        try {
+            if (Directory.Exists(Path.GetDirectoryName(filePath)) && _serializableBreakpointsHolder != null) {
```
codengine: An early return would improve readability and reduce nesting depth.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/ProgramSerializedBreakpoints.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Spice86.Shared.Emulator.VM.Breakpoint.Serializable;
+
+/// <summary>
+/// Represents a collection of breakpoints associated with a specific program, identified by its SHA256 hash.
+/// </summary>
+public record ProgramSerializedBreakpoints {
```
codengine: The record name is a bit misleading. I would call it ProgramSerializableBreakpoints

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System.Text.Json.Serialization;
+
+/// <summary>
+/// Base class for serializable breakpoint data.
+/// </summary>
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
+[JsonDerivedType(typeof(SerializedBreakpoint), typeDiscriminator: "breakpoint")]
+[JsonDerivedType(typeof(SerializedBreakpointRange), typeDiscriminator: "range")]
+public record SerializedBreakpoint {
```
codengine: The record name is a bit misleading. I would call it SerializableBreakpoint

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
codengine: I think its better to put the version onto the container (ProgramSerializableBreakpoints). I see no point to do that per breakpoint.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpointRange.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Spice86.Shared.Emulator.VM.Breakpoint.Serializable;
+/// <summary>
+/// Class for serializing memory range breakpoints
+/// </summary>
+public record SerializedBreakpointRange : SerializedBreakpoint {
```
codengine: The record name is a bit misleading. I would call it SerializableBreakpointRange

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoints.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Spice86.Shared.Emulator.VM.Breakpoint.Serializable;
+/// <summary>
+/// Container class for a collection of serialized breakpoints.
+/// </summary>
+public record SerializedBreakpoints {
```
codengine: The record name is a bit misleading. I would call it SerializableBreakpoints

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoints.cs
```diff
@@ -0,0 +1,15 @@
...
+    /// <summary>
+    /// Gets the format version.
+    /// </summary>
+    public int Version { get; init; } = 1;
```
codengine: Same here. I would put the version on the container.

src/Spice86.Shared/Interfaces/ISerializableBreakpointsHolder.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Spice86.Shared.Interfaces;
+
+using Spice86.Shared.Emulator.VM.Breakpoint.Serializable;
+
+/// <summary>
+/// This interface defines a method for generating a serializable object that encapsulates all breakpoints of the internal debugger
+/// </summary>
+public interface ISerializableBreakpointsHolder {
```
codengine: A generic comment (and a bit personal taste): I see it as an anti pattern to use interfaces if there is only one implementation (and even more so if this is going to stay that way).

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/EmulatorBreakpointsSerializer.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        // If no hash is specified,compute it ourselves
+        if (string.IsNullOrWhiteSpace(configuration.Exe) || !File.Exists(configuration.Exe)) {
+            throw new FileNotFoundException(configuration.Exe);
+        }
+        return ConvertUtils.ByteArrayToHexString(SHA256.HashData(File.ReadAllBytes(configuration.Exe)));
+    }
+
+    public void AddSerializableBreakpointsHolder(ISerializableBreakpointsHolder serializableBreakpointsHolder) {
```
codengine: The method does not "add" a breakpoint holder. Why isn't it initialized from the constructor?

codengine (changes requested):

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
maximilien-noal: I've removed this field.

src/Spice86/ViewModels/BreakpointsViewModel.cs
maximilien-noal: It is more the responsibility of the EmulatorBreakpointsManager now, to create and load/save user breakpoints.

src/Spice86.Core/Emulator/Function/Dump/EmulatorStateSerializer.cs
maximilien-noal: The path is passed now.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/ProgramSerializedBreakpoints.cs
maximilien-noal: Done.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
maximilien-noal: Done.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoint.cs
maximilien-noal: I removed the version stuff.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoints.cs
maximilien-noal: Done.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpoints.cs
maximilien-noal: I removed it.

src/Spice86.Shared/Interfaces/ISerializableBreakpointsHolder.cs
maximilien-noal: It permits interface separation - no need to depend on an entire class when you only want a single method.

It can be expanded later to the GDB server so it can have its breakpoints serialized/restored too.

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/EmulatorBreakpointsSerializer.cs
maximilien-noal: Removed - we use ctor dependency injection now.

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/EmulatorBreakpointsSerializer.cs
maximilien-noal: There's no folder generation, we simply pass the dirPath from EmulatorStateSerializer now.

The SHA256 hash is always created, as suggested.

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/EmulatorBreakpointsSerializer.cs
maximilien-noal: Done.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializedBreakpointRange.cs
maximilien-noal: Done.

src/Spice86.Shared/Emulator/VM/Breakpoint/Serializable/SerializableUserBreakpoint.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Text.Json.Serialization;
+
+/// <summary>
+/// Base class for serializable breakpoint data created by the user in the internal Spice86 debugger.
+/// </summary>
+/// <remarks>Not everything is serialized, this is why this is a different set of classes.</remarks>
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
+[JsonDerivedType(typeof(SerializableUserBreakpoint), typeDiscriminator: "breakpoint")]
+[JsonDerivedType(typeof(SerializableUserBreakpointRange), typeDiscriminator: "range")]
+public record SerializableUserBreakpoint {
```
kevinferrare, line 9: I would remove User or Serializable from the name, it's redundant, unless there can be serializable breakpoints that are not user :)

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/UserBreakpointsSerializer.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Security.Cryptography;
+
+/// <summary>
+/// Serializes (on startup) and deserialize (on app shutdown) breakpoints created by the user in the internal Spice86 debugger UI.
+/// </summary>
+public class UserBreakpointsSerializer {
+    private const string BreakpointsFileNameFormat = "Breakpoints_{0}.json";
+    private readonly ILoggerService _loggerService;
+    private readonly ISerializableBreakpointsSource _serializableBreakpointsSource;
+    private readonly string _programHash;
```
kevinferrare: I think handling the hash and so on shouldnt be this class responsibility.
Code here should just dump the breakpoints in a json file without caring about the hash, and the json file should be in the same folder as the other dumped data.
The dump folder should contain the hash of the exe if it is not provided by the config parameter.
This will improve dumps and make things better for the future.

src/Spice86.Core/Emulator/VM/Breakpoint/AddressRangeBreakPoint.cs
```diff
@@ -0,0 +1,40 @@
+namespace Spice86.Core.Emulator.VM.Breakpoint;
+
+using Spice86.Shared.Emulator.VM.Breakpoint;
+
+using System;
+
+/// <summary>
+/// Represents a breakpoint that is triggered when the CPU's execution address is within a specified range.
+/// </summary>
+public class AddressRangeBreakPoint : AddressBreakPoint {
```
kevinferrare: Is this working?

The breakpoints are actually mapped in a dictionary with 1 key per address, so that the emulator doesnt loop over them over and over (would wreck perfs).

So for this to work, this would need to be injected for several keys and I dont think this is currently done.

Maybe this could be left aside for another feature? By the way, is the feature working from the UI??

maximilien-noal: I tested the PR entirely before submitting it again.

The code might be not nice but range breakpoints and all the other kind of breakpoints do work.

maximilien-noal: class removed.

src/Spice86.Core/Emulator/VM/Breakpoint/Serializable/UserBreakpointsSerializer.cs
maximilien-noal: This has been done.

src/Spice86.Core/Emulator/VM/Breakpoint/EmulatorBreakpointsManager.cs
```diff
@@ -115,4 +124,125 @@ public void CheckExecutionBreakPoints() {
...
+            BreakPointType type = group.Key.BreakPointType;
+
+            yield return CreateSerializableMemoryBreakpoint(rangeStart, rangeEnd, type);
+        }
+    }
+
+    private static SerializableUserBreakpoint CreateSerializableMemoryBreakpoint(
+        long start, long end, BreakPointType type) {
+        if (start == end) {
+            return new SerializableUserBreakpoint {
```
kevinferrare: Wouldnt it make sense to have only one type with start / end? if start == end, this would mean single breakpoint

src/Spice86.Core/Emulator/VM/Breakpoint/EmulatorBreakpointsManager.cs
```diff
@@ -115,4 +124,125 @@ public void CheckExecutionBreakPoints() {
...
+
+        if (!memoryBreakpoints.Any()) {
+            yield break;
+        }
+
+        foreach (var group in memoryBreakpoints
+            .Select((bp, index) => new {
+                Breakpoint = bp,
+                Index = index,
+                GroupKey = bp.Address - index
```
kevinferrare: Are you sure this works if input list is not sorted? I would think grouping logic depends on address ordering is the same as index ordering
Consecutive addresses should have consecutive indexes

xorangekiller:
I tried it out, and it works pretty well. My only minor complaint is that disabled breakpoints are not serialized. It would be nice to be able to save them too so that they can be reenabled later.
```
$ cat FB917AD5E46605EEC58A5C0608A1FD45DDA9BCBF00432D5BDBA71E0EBCA62821/Breakpoints.json | jq .
{
  "ProgramHash": "FB917AD5E46605EEC58A5C0608A1FD45DDA9BCBF00432D5BDBA71E0EBCA62821",
  "SerializedBreakpoints": {
    "Breakpoints": [
      {
        "$type": "breakpoint",
        "Trigger": 29436,
        "Type": 0,
        "IsEnabled": true
      },
      {
        "$type": "breakpoint",
        "Trigger": 28855,
        "Type": 0,
        "IsEnabled": true
      }
    ]
  }
}
```

src/Spice86.Core/Emulator/VM/Breakpoint/EmulatorBreakpointsManager.cs
maximilien-noal: this has been done.

src/Spice86.Core/Emulator/VM/Breakpoint/EmulatorBreakpointsManager.cs
maximilien-noal: This code has been replaced/reworked.

maximilien-noal:
> I tried it out, and it works pretty well. My only minor complaint is that disabled breakpoints are not serialized. It would be nice to be able to save them too so that they can be reenabled later.
> 
> ```
> $ cat FB917AD5E46605EEC58A5C0608A1FD45DDA9BCBF00432D5BDBA71E0EBCA62821/Breakpoints.json | jq .
> {
>   "ProgramHash": "FB917AD5E46605EEC58A5C0608A1FD45DDA9BCBF00432D5BDBA71E0EBCA62821",
>   "SerializedBreakpoints": {
>     "Breakpoints": [
>       {
>         "$type": "breakpoint",
>         "Trigger": 29436,
>         "Type": 0,
>         "IsEnabled": true
>       },
>       {
>         "$type": "breakpoint",
>         "Trigger": 28855,
>         "Type": 0,
>         "IsEnabled": true
>       }
>     ]
>   }
> }
> ```

thanks for the review, this is fixed! :)

kevinferrare (approved):

PR #1307, opened 2025-09-05
feat: call interrupt 0x1C on each timer tick

codengine:
### Description of Changes
Call Int1C on each timer tick to allow timer-specific hooks. This fixes Exodus.
The old CPU is currently broken with this change, thats why its a draft.

### Rationale behind Changes
Calling the Int 1C is according to spec.

### DIFF
--- a/tests/Spice86.Tests/MachineTest.cs
+++ b/tests/Spice86.Tests/MachineTest.cs
@@ -1,5 +1,7 @@
 namespace Spice86.Tests;
 
+using FluentAssertions;
+
 using JetBrains.Annotations;
 
 using Serilog;
@@ -9,8 +11,6 @@ using Spice86.Core.Emulator.CPU.CfgCpu.Feeder;
 using Spice86.Core.Emulator.CPU.CfgCpu.ParsedInstruction;
 using Spice86.Core.Emulator.CPU.CfgCpu.ParsedInstruction.Instructions;
 using Spice86.Core.Emulator.Errors;
-using Spice86.Core.Emulator.InterruptHandlers.Common.MemoryWriter;
-using Spice86.Core.Emulator.InterruptHandlers.Common.RoutineInstall;
 using Spice86.Core.Emulator.IOPorts;
 using Spice86.Core.Emulator.Memory;
 using Spice86.Core.Emulator.VM;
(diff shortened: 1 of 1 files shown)


codengine:
@kevinferrare @maximilien-noal Please re-approve.

I had to fix a small thing and also rebased and squashed + co-authored Kevin.

codengine:
@kevinferrare Could you also please have a look at the failing test?

kevinferrare:
Test is failing because code is not doing the correct thing here

src/Spice86.Core/Emulator/InterruptHandlers/Timer/DummyInt1CHandler.cs
maximilien-noal, line 1: This should be either implemented, or not implemented.

If it is not implemented, then an exception is fine.

If there is an implementation, it should be fully implemented.

src/Spice86.Core/Emulator/InterruptHandlers/Timer/TimerInt8Handler.cs
```diff
@@ -42,6 +43,7 @@ public TimerInt8Handler(IMemory memory, IFunctionHandlerProvider functionHandler
     /// <inheritdoc />
     public override void Run() {
         TickCounterValue++;
+        _emulationLoopRecall.RunInterrupt(0x1C);
```
maximilien-noal: I'm not fond of emulationloopRecall, which is essentialy a hack inherited from DOSBox that I implemented.

Why not just call the INT1C Run() method directly ?

maximilien-noal (changes requested):
A few requested changes on my end - for now.

I think the unit test approach is great!

While this fixes EXODUS, is it supposed to fix Dune floppy freeze too ?

I tested with Dune floppy and it still freezes.

src/Spice86.Core/Emulator/InterruptHandlers/Timer/DummyInt1CHandler.cs
codengine: The "normal" implementation is that Int 1C is basically just an IRET, which the dummy handler represents.
We could also catch the exception that is thrown if no user int handler is defined for that interrupt, but first: i dont like exceptions for normal control flow and second, it doesnt model how real dos implements it.

codengine: It doesn't fix Dune, there is another reason why it freezes. But calling software-defined Int 1C overrides on each timer tick is definitely the right thing to do and will probably many issues with other games as well.

maximilien-noal (approved):

Issue #1290, opened 2025-08-22
Cadaver freezes on master with CFGCPU enabled (GAME.EXE)


maximilien-noal:

maximilien-noal:
<img width="1287" height="1002" alt="Image" src="https://github.com/user-attachments/assets/1a2cd85d-9f0b-4ca3-ab61-86d3ff6994f5" />

maximilien-noal:
Works fine with the old CPU.

codengine:
Fixed in https://github.com/OpenRakis/Spice86/pull/1309

PR #1309, opened 2025-09-06
fix: cadaver CfgCpu hanging

codengine:
### Description of Changes
This PR avoid busy-waiting for keyboard events and solves #1290. CfgCpu chokes on it.

### Rationale behind Changes
Busy waiting is the wrong thing to do while waiting for Keystrokes, though we have to block the instruction. The best way to solve it IMHO is to set a flag that signals "please stay at the current instruction, process interrupts, dma etc. and come back".

The benefit is that the code is much more responsive and reacts immediately to keystrokes while also consuming much less CPU resources with CfgCpu.

I was not able to implement the same hack in the old CPU, therefore I added a compatibility mode where it would still use the busy wait with the old CPU.

It is a bit hacky but it works :)

DosBox does it similarly.

### Suggested Testing Steps
Run the game, select a language.

### DIFF
--- a/src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
+++ b/src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
@@ -13,8 +13,6 @@ using Spice86.Shared.Interfaces;
 
 using System.Diagnostics.CodeAnalysis;
 
-using static System.Runtime.CompilerServices.RuntimeHelpers;
-
 /// <summary>
 /// The keyboard controller interrupt (INT16H)
 /// </summary>
@@ -42,7 +40,6 @@ public class KeyboardInt16Handler : InterruptHandler {
         _emulationLoopRecall = emulationLoopRecall;
         _biosDataArea = biosDataArea;
         _biosKeyboardBuffer = biosKeyboardBuffer;
-        AddAction(0x00, GetKeystroke);
         AddAction(0x01, () => GetKeystrokeStatus(true));
         AddAction(0x02, GetShiftFlags);
         AddAction(0x1D, () => Unsupported(0x1D));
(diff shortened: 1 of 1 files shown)


maximilien-noal (changes requested):
Instead of EmulationLoopRecall (I authored it, but I still think it(s hacky) or modifiying the CPU, which is indeed hacky, the real fix seems to be what the actual hardware does:

In a real BIOS implementation, INT 16H function 0 waits for a key to be pressed before it returns.

See the seabios BIOS as an example :

https://github.com/OpenRakis/SeaBIOS/blob/master/src/kbd.c

handle_1600 -> dequeue_key ->

`    

    static void
    dequeue_key(struct bregs *regs, int incr, int extended)
    {
    yield();
    u16 buffer_head;
    u16 buffer_tail;
    for (;;) {
        buffer_head = GET_BDA(kbd_buf_head);
        buffer_tail = GET_BDA(kbd_buf_tail);

        if (buffer_head != buffer_tail)
            break;
        if (!incr) {
            regs->flags |= F_ZF;
            return;
        }
        yield_toirq();
    }

    u8 ascii_code = GET_FARVAR(SEG_BDA, *(u8*)(buffer_head+0));
    u8 scan_code  = GET_FARVAR(SEG_BDA, *(u8*)(buffer_head+1));
    if ((ascii_code == 0xF0 && scan_code != 0)
        || (ascii_code == 0xE0 && !extended))
        ascii_code = 0;
    regs->ax = (scan_code << 8) | ascii_code;

    if (!incr) {
        regs->flags &= ~F_ZF;
        return;
    }
    u16 buffer_start = GET_BDA(kbd_buf_start_offset);
    u16 buffer_end   = GET_BDA(kbd_buf_end_offset);

    buffer_head += 2;
    if (buffer_head >= buffer_end)
        buffer_head = buffer_start;
    SET_BDA(kbd_buf_head, buffer_head);
    }

`

busy wait mixed with yield to give back control temporarely.

There's also an ASM example here (PC/XT BIOS) :

https://github.com/virtualxt/pcxtbios/blob/master/pcxtbios.asm

```
int_16_entry:

	sti					; Keyboard BIOS services
	push	ds
	push	bx

	mov	bx, 40h
	mov	ds, bx				; Load work segment

	cmp	ah, 5
	je	@@stuff				; Stuff keyboard buffer, ah=5

	mov	bx, ax				; Save function number to check for
						;   extended call later

	and	ah, 0Fh				; Translate enhanced keyboard function calls
						; ah=10h/11h/12h -> ah=00h/01h/02h
	or	ah, ah
	jz	@@read				; Read keyboard buffer, ah=0
	dec	ah
	jz	@@check				; Set Z if char ready, ah=1
	dec	ah
	jz	@@shift				; Return shift in al, ah=2

@@end:
	pop	bx				; Exit INT_16 keyboard service
	pop	ds
	iret

@@read:
	cli					; No interrupts, alters buffer
	mov	ax, [ds:1Ah]			;   point to buffer head
	cmp	ax, [ds:1Ch]			; If not equal to buffer tail
	jnz	@@have_char			;   char waiting to be read
	sti					; Else allow interrupts
	jmp	@@read				;   wait for him to type

@@have_char:
	test	bh, 10h				; Test for extended function call
	pushf					; Save zf for later

	xchg	ax, bx
	mov	ax, [bx]			; Fetch the character

	popf					; Is this an extended function call?
	jnz	@@no_translation		; Yes so don't change extended scan codes

	cmp	al, 0E0h			; Is scan code E0h?
	jne	@@no_translation
	xor	al, al				; If so translate to 00h for standard function
@@no_translation:

	inc	bx				; Point to next character
	inc	bx				;   char = scan code + shift
	mov	[ds:1Ah], bx			; Save position in head
	cmp	bx, [ds:82h]			;   buffer overflowed?
	jnz	@@end				;   no, done
	mov	bx, [ds:80h]			; Else reset to point at start
	mov	[ds:1Ah], bx			;   and correct head position
	jmp	short @@end

endp	int_16

```

function 0 (@@read) also uses busy waiting.

This is also expected behavior by the DOS console driver (otherwise its while loop doesn't behave properly).

As for DOSBox, it esentially disallows interrupts in INT9H (callback.cpp - OK, fine), then in INT9H it reads the keyboard port through CPU register AL (which is not how real hardware reads I/O ports - reading ports should not involve the CPU), and to do *that* , their IO port dispatcher recalls the emulation loop (same hack I implemented, although I do not like it).

Which is how their INT9H get its in AL, seemingly out of nowhere at first:


```

/* the scancode is in reg_al */
static Bitu IRQ1_Handler(void) {
/* handling of the locks key is difficult as sdl only gives
 * states for numlock capslock. 
 */
	const auto scancode = reg_al; // Read the code

```

bios_keyboard.cpp both defines their INT16H and INT9H (IRQ1) handlers.

maximilien-noal:
Tested with many games, including Dune, Krondor, Dune II, Monkey Island 1, Wolf3D, and Shangai II, without issues.

Approved! This a very nice fix, and a very important one.

Your help is very much appreciated @codengine  - I wanted to fix this part of the code for so long.

Not only it's the proper way of doing it, but it also avoids all the dangers and pitfalls of the EmulationLoopRecall class (CFGCPU graph messed up, old CPU parser confused).

Approved!

maximilien-noal (approved):

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
```diff
@@ -62,8 +59,89 @@ private void Unsupported(int operation) {
...
+    ///     L_HANDLER:
+    ///     cmp ah, 00h
+    ///     jz  L_AH00
+    /// 
+    ///     L_DEFAULT:
+    ///     int 0A4h   ; DefaultDispatch -> calls C# Run()
+    ///     iret
+    /// 
+    ///     L_AH00:
+    ///     int 0A0h   ; CallbackHasKey -> sets ZF=0 if a key is present
```
kevinferrare, line 77: it's not an int, it's a callback, this is a bit confusing

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
```diff
@@ -62,8 +59,89 @@ private void Unsupported(int operation) {
...
+    ///     jnz have_key
```
kevinferrare, line 78: jnz have_key => so it will jump if ZF = 1 which means if "_biosKeyboardBuffer.IsEmpty"

Is the label right? Is it me interpreting that wrongly?

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
```diff
@@ -62,8 +59,89 @@ private void Unsupported(int operation) {
...
+        // L_AH00:
+        // callback HasKey (sets ZF=0 when key present)
+        memoryAsmWriter.RegisterAndWriteCallback(0xA0, CallbackHasKey);
+        // JNZ have_key (+0x04 to skip INT 09h and the backward jump)
+        memoryAsmWriter.WriteUInt8(0x75);
+        memoryAsmWriter.WriteUInt8(0x04);
+        // INT 09h
+        memoryAsmWriter.WriteInt(0x09);
+        // JMP short back to L_AH00 (-0x09)
+        memoryAsmWriter.WriteUInt8(0xEB);
```
kevinferrare, line 123: Why not adding some opcodes in memoryAsmWriter as reuseable functions instead of writing hexadecimal directly?
For offsets of jumps, they could be computed from the diff between the address of the writer at label time and after jump? (the jump function could take an absolute segmented address and compute the diff?)

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
codengine: The callback is dispatched through an int, no?

src/Spice86.Core/Emulator/InterruptHandlers/Input/Keyboard/KeyboardInt16Handler.cs
codengine: Good point, I'll have a look later

PR #1312, opened 2025-09-07
DOS Function INT21H 0x4B 'LOAD and/or EXEC'

maximilien-noal:
Preview PR for DOS process support without command.com

The emulated program started with Configuration.Exe points to itself so we detect it properly as the first program when exiting.

Fun stuff!

image

That's Lemmings II which stopped outputting a memory error.

Doom8088 loops forever now.

image

Zeliard runs already.

Progress....

### DIFF
--- a/.github/workflows/nuget.yml
+++ b/.github/workflows/nuget.yml
@@ -30,29 +30,29 @@ jobs:
 
     - name: Package NuGet
       working-directory: ./src
-      run: dotnet pack --include-symbols --include-source --configuration Release
+      run: dotnet pack --include-symbols --include-source --configuration Release -p:NoWarn=NU5104
 
     - name: Upload NuGet Bufdio.Spice86
-      if: github.ref == 'refs/heads/master'
+      if: github.ref == 'refs/heads/master' && matrix.os == 'ubuntu-latest'
       working-directory: ./src/Bufdio.Spice86/bin/Release
-      run: dotnet nuget push Bufdio.Spice86.10.0.0.nupkg ${{secrets.NUGET_API_KEY}} -Source 'https://api.nuget.org/v3/index.json' -SkipDuplicate
+      run: dotnet nuget push Bufdio.Spice86.11.0.0.nupkg --api-key "${{ secrets.NUGET_API_KEY }}" --source "https://api.nuget.org/v3/index.json" --skip-duplicate
 
     - name: Upload NuGet Spice86.Shared
-      if: github.ref == 'refs/heads/master'
+      if: github.ref == 'refs/heads/master' && matrix.os == 'ubuntu-latest'
       working-directory: ./src/Spice86.Shared/bin/Release
-      run: dotnet nuget push Spice86.Shared.10.0.0.nupkg ${{secrets.NUGET_API_KEY}} -Source 'https://api.nuget.org/v3/index.json' -SkipDuplicate
+      run: dotnet nuget push Spice86.Shared.11.0.0.nupkg --api-key "${{ secrets.NUGET_API_KEY }}" --source "https://api.nuget.org/v3/index.json" --skip-duplicate
 
     - name: Upload NuGet Spice86.Logging
-      if: github.ref == 'refs/heads/master'
+      if: github.ref == 'refs/heads/master' && matrix.os == 'ubuntu-latest'
       working-directory: ./src/Spice86.Logging/bin/Release
-      run: dotnet nuget push Spice86.Logging.10.0.0.nupkg ${{secrets.NUGET_API_KEY}} -Source 'https://api.nuget.org/v3/index.json' -SkipDuplicate
+      run: dotnet nuget push Spice86.Logging.11.0.0.nupkg --api-key "${{ secrets.NUGET_API_KEY }}" --source "https://api.nuget.org/v3/index.json" --skip-duplicate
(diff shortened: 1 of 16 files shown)


maximilien-noal:
@copilot following on my footsteps, ensure that we properly support and implement:

COMMAND.COM on the Z drive - not interactive
BATCH support from COMMAND.COM - unit tested
CP/M File Control Block based APIs
All modes of executions of LOAD AND OR EXEC and subfunctions
Implement TSR support
FULL BIOS services (all interrupts, including time and disk interrupts)

All integration tested with asembly program tests that will certainly fail first, but will guide too.

All methods should be public, documented, high level - with enum and MemoryBasedDataStructures (like DosExecParamBlock, already done this part I think) and also unit tested

Documentation and sources to follow:

MS-DOS 4 sources on github for EXEC.ASM

DOS TECHNICAL INFORMATION
 Programming Technical Reference - IBM
 Copyright 1988, Dave Williams

The MS-DOS Encyclopedia
Section II: Programming in the MS-DOS Environment

https://www.pcjs.org/documents/books/mspl13/msdos/encyclopedia/section2/

Writing Memory Resident software:

https://faculty.kfupm.edu.sa/coe/mudawar/coe205/manuals/MASMProgrGuide/11-MemoryResident.pdf

This information about TSRs:

Chapter 18
Page 1034
Assuming you’ve initialized the Int10InUse variable to zero, the in use flag will contain zero when it is
safe to execute an int 10h instruction in your TSR, it will contain a non-zero value when the interrupt 10h
handler is busy. You can use this flag like the InDOS flag to defer the execution of your TSR code.
Like DOS, there are certain BIOS routines that may take an indefinite amount of time to complete.
Reading a key from the keyboard buffer, reading or writing characters on the serial port, or printing char-
acters to the printer are some examples. While, in some cases, it is possible to create a wrapper that lets
your TSR activate itself while a BIOS routine is executing one of these polling loops, there is probably no
benefit to doing so. For example, if an application program is waiting for the printer to take a character
before it sends another to printer, having your TSR preempt this and attempt to send a character to the
printer won’t accomplish much (other than scramble the data sent to the print). Therefore, BIOS wrappers
generally don’t worry about indefinite postponement in a BIOS routine.
5, 8, 9, D, E, 10, 13, 16, 17, 21, 28
If you run into problems with your TSR code and certain application programs, you may want to
place wrappers around the following interrupts to see if this solves your problem: int 5, int 8, int 9, int B,
int C, int D, int E, int 10, int 13, int 14, int 16, or int 17. These are common culprits when TSR problems
develop.
18.3.3 Reentrancy Problems with Other Code
Reentrancy problems occur in other code you might call as well. For example, consider the UCR Stan-
dard Library. The UCR Standard Library is not reentrant. This usually isn’t much of a problem for a couple
of reasons. First, most TSRs do not call Standard Library subroutines. Instead, they provide results that
normal applications can use; those applications use the Standard Library routines to manipulate such
results. A second reason is that were you to include some Standard Library routines in a TSR, the applica-
tion would have a separate copy of the library routines. The TSR might execute an strcmp instruction
while the application is in the middle of an strcmp routine, but these are not the same routines! The TSR is
not reentering the application’s code, it is executing a separate routine.
However, many of the Standard Library functions make DOS or BIOS calls. Such calls do not check to
see if DOS or BIOS is already active. Therefore, calling many Standard Library routines from within a TSR
may cause you to reenter DOS or BIOS.
One situation does exist where a TSR could reenter a Standard Library routine. Suppose your TSR has
both passive and active components. If the main application makes a call to a passive routine in your TSR
and that routine call a Standard Library routine, there is the possibility that a system interrupt could inter-
rupt the Standard Library routine and the active portion of the TSR reenter that same code. Although such
a situation would be extremely rare, you should be aware of this possibility.
Of course, the best solution is to avoid using the Standard Library within your TSRs. If for no other
reason, the Standard Library routines are quite large and TSRs should be as small as possible.

Writing DOS Device drivers in C

https://ftpmirror.your.org/pub/misc/bitsavers/pdf/microsoft/msdos_4.0/Adams_-_Writing_DOS_Device_Drivers_in_C_1990.pdf

also do the same for our DOS IOCTL and DOS Memory Manager and DOS File Manager implementations.

Assembly tests, unit tests, full compatibility test suite first of:

BIOS
DOS IOCTL
COMMAND.COM
BATCH support
DOS PROCESS support
DOS TSR support
DOS overlay support
DOS codepage support (we use a Nuget for that, but it's barely implemented)
DOS DBCS support and structure initialization
FULL DOS INT21H and other DOS INTs services.
Ctrl^C and Ctrl^Break support
DOS MCB to protect the DOS Driver area and one also to protect the DOS PSP ENV Block.

For any memory based structure, usr a MemoryBasedDataStructure based class.

This is massive - but with tests, sources, and documentation, our DOS and BIOS has a real chance to be complete and with more code quality and more stability.

AUOEXEC.BAT should be created from start Configuration. The C folder mount point should be resolved once on AUTOEXEC.BAT generation.

CONFIG.SYS is ignored - our DOS does not depend on CONFIG.SYS and that's good.

Things to ignore too:

CD support, MSCDEX support, and floppy drives support

One thing not to redo:

XMS and EMS - they are perfect.

One other thing to ignore: the RTC CLock and related DOS and BIOS functions - this is another PR.

PR #1318, opened 2025-09-10
fix: some minor mouse-related fixes and refactorings

codengine:
### Description of Changes
Some minor refactorings. There are five categories of changes:
- avoid local variable allocation of mouse buttons
- Documentation fix
- Safe inspection fixes
- Safety measures
- Real bugs (SetMouseMickeyPixelRatio, GetLastPressedY, GetLastReleasedY)

Regarding the real bugs:

**SetMouseMickeyPixelRatio**
I read the documentation and implementation and it should be divided by 8 instead of multiplied by 8.

**GetLastPressedY** & **GetLastReleasedY**
Both used LastPressedX instead of LastPressedY

### Rationale behind Changes
Safety, correctness

### DIFF
--- a/src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/IMouseDevice.cs
+++ b/src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/IMouseDevice.cs
@@ -48,10 +48,10 @@ public interface IMouseDevice : IIOPortHandler {
     MouseEventMask LastTrigger { get; }
 
     /// <summary>
-    ///     The sample rate of the mouse.
+    ///     The sample rate of the mouse. Currently unused.
     /// </summary>
     int SampleRate { get; set; }
-
+    
     /// <summary>
     ///     The amount of movement in the Y direction since the last update.
     /// </summary>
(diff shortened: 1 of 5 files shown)


src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/IMouseDevice.cs
```diff
@@ -47,11 +47,6 @@ public interface IMouseDevice : IIOPortHandler {
     /// </summary>
     MouseEventMask LastTrigger { get; }
 
-    /// <summary>
```
maximilien-noal, line 50: It's important to keep this - as the sample rate is supposed to be configurable.

`Setting the Interrupt Rate for the InPort
Mouse
If you are using an InPort Mouse, you can use one of the
following command line switch settings to specify the
in terrupt rate for the mouse:
Switch Setting
/RO
/Rl
/R2
/R3
/R4
Interrupt Rate
disabled
30Hz (default)
50Hz
100Hz
200Hz`

[MS-DOS Mouse driver reference](https://archive.org/details/bitsavers_microsoftm1186MicrosoftMouseProgrammersReferenceGu_6193519)

maximilien-noal (changes requested):
Works fine with Krondor.

There's just a small requested change to keep the Sample Rate in the IMouseDevice interface.

src/Spice86.Core/Emulator/InterruptHandlers/Input/Mouse/IMouseDevice.cs
codengine: @maximilien-noal I removed it because it was neither set nor read. Should it still be kept?

maximilien-noal: Yep it was there for future work

codengine: I added it back and force pushed the changes

PR #1338, opened 2025-09-14
Put back callback number on 16 bits

kevinferrare:
This is a prerequisite for callback allocation.


maximilien-noal:
Tested with Dune II and XMS => OK.

PR #1373, opened 2025-10-07
Feature/rtc clock and bios functions

maximilien-noal:
Preview PR that should be done soon enough for the RTC Clock and related DOS INT21H and BIOS tests and services.

It is a small PR to see if we can have equivalents of a small subset of timed events from DOSBox Staging but for the RTC / CMOS.

The keyboard and Adlib Gold and OPL2 and Sound Blaster emulation all use timed events based on emulated Ticks (see pic.cpp)

And this has been my source of problems and show stoppers for a long while in a lot of subjects...


maximilien-noal:
@copilot implement the RTC Clock and associated BIOS functions.

Replace the Clock class.

Removed the try/catch approach.

Implement the related BIOS wait function.

Implement the related DOS $Clock device.

Remove the Clock in the DOS constructor.

The RTC Clock and CMOS classes should be public, fully documented, unit tested with ASM tests and unit tests (ASM tests for emulation accuracy, and unit tests for guidance about specific units of features)

PR #1387, opened 2025-10-08
fix: segmented addressing boundary, using tests

maximilien-noal:
### Description of Changes

Fixes memory read/write problem around segment (as in segmented memory addressing) boudaries

### Rationale behind Changes

New unit tests were introduced first that revealed the issue.

Then memory indexers were eventually fixed.

For a brief moment INT2F tests for XMS were broken by the changes (and PoP was too) because callback numbers were changed from (for example) 0x8 to 0x8000 (the latter doesn't exist -> exception) - but this was fixed quickly.

### Suggested Testing Steps

Already tested quickly with:

Dune CD
Dune Floppy
Another World
Betrayal at Krondor
B.A.T. II (but as with the master branch, it needs an unimplemented interrupt function)
Lost Eden
Prince of Persia
Dune 2
Cruise for a Corpse
Wolfenstein 3D
Hocus Pocus
Jill of the Jungle
Enviro-Kids
Blake Stone: Aliens of Gold
Oh No! More Lemmings

All only with the CFGCPU

## Performance

After testing, despite the more complicated operations, performance is the same as master.

That is, between 14M aand 18M instructions per second on my machine.

## Does it fix the Dune CFGCPU + SB PCM bug ?

Nope. :/

### DIFF
--- a/src/Spice86.Core/Emulator/Memory/Indexer/MemoryIndexer.cs
+++ b/src/Spice86.Core/Emulator/Memory/Indexer/MemoryIndexer.cs
@@ -21,9 +21,9 @@ public abstract class MemoryIndexer<T> : Indexer<T>, IList<T> {
     /// Gets or sets the data at the specified segmented address and offset in the memory.
     /// </summary>
     /// <param name="address">Segmented address at which to access the data</param>
-    public abstract T this[SegmentedAddress address] {
-        get;
-        set;
+    public T this[SegmentedAddress address] {
+        get => this[address.Segment, address.Offset];
+        set => this[address.Segment, address.Offset] = value;
     }
 
     /// <inheritdoc />
--- a/src/Spice86.Core/Emulator/Memory/Indexer/UInt8Indexer.cs
+++ b/src/Spice86.Core/Emulator/Memory/Indexer/UInt8Indexer.cs
@@ -32,15 +32,6 @@ public class UInt8Indexer : MemoryIndexer<byte> {
         set => _byteReaderWriter[MemoryUtils.ToPhysicalAddress(segment, offset)] = value;
     }
 
-    /// <summary>
-    /// Gets or sets the data at the specified segmented address and offset in the memory.
-    /// </summary>
-    /// <param name="address">Segmented address at which to access the data</param>
-    public override byte this[SegmentedAddress address] {
-        get => this[address.Segment, address.Offset];
-        set => this[address.Segment, address.Offset] = value;
-    }
-
     /// <inheritdoc/>
     public override int Count => _byteReaderWriter.Length;
 }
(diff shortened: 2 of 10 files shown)


src/Spice86.Core/Emulator/Memory/Indexer/UInt8Indexer.cs
```diff
@@ -20,6 +22,25 @@ public override byte this[uint address] {
         set => _byteReaderWriter[address] = value;
     }
 
+    /// <summary>
+    /// Gets or sets the data at the specified segment and offset in the memory.
+    /// </summary>
+    /// <param name="segment">The segment of the element to get or set.</param>
+    /// <param name="offset">The offset of the element to get or set.</param>
+    public new byte this[ushort segment, ushort offset] {
```
kevinferrare: Why not making the base methods virtual instead of hiding them?

maximilien-noal: It's done now, good catch! :)

src/Spice86.Core/Emulator/Memory/Indexer/MemoryIndexer.cs
```diff
@@ -12,7 +12,7 @@ public abstract class MemoryIndexer<T> : Indexer<T>, IList<T> {
     /// </summary>
     /// <param name="segment">The segment of the element to get or set.</param>
     /// <param name="offset">The offset of the element to get or set.</param>
-    public T this[ushort segment, ushort offset] {
+    public virtual T this[ushort segment, ushort offset] {
```
kevinferrare: Should be abstract. Using physical address is wrong

maximilien-noal: done. For both set of get/set segmented addressing accessors

src/Spice86.Core/Emulator/Memory/Indexer/MemoryIndexer.cs
```diff
@@ -12,18 +12,18 @@ public abstract class MemoryIndexer<T> : Indexer<T>, IList<T> {
...
     }
 
     /// <summary>
     /// Gets or sets the data at the specified segmented address and offset in the memory.
     /// </summary>
     /// <param name="address">Segmented address at which to access the data</param>
-    public T this[SegmentedAddress address] {
-        get => this[address.Segment, address.Offset];
-        set => this[address.Segment, address.Offset] = value;
+    public abstract T this[SegmentedAddress address] {
```
kevinferrare: Why not keeping this? This would avoid duplicating logic twice, this is supposed to be syntax sugar :)

maximilien-noal: Done - and only kept in MemoryIndexer. It was indeed copy/pasted everywhere.

PR #1406, opened 2025-10-16
Fix/UI and emulator threads race condition

maximilien-noal:
fixes race condition between UI thread (that sends mouse/keyboard events) and emulator thread by using an input event queue.

It's always really small, so perf impact is not noticeable at all.

DOSBox does the same, but only because it's the same solution to a common problem.

### DIFF
--- a/src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
+++ b/src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
@@ -98,8 +98,8 @@ public sealed class Keyboard : DefaultIOPortHandler {
             case KeyboardPorts.Data:
                 _a20Gate.IsEnabled = Command switch {
                     KeyboardCommand.WriteOutputPort => (value & 2) > 0,
-                    KeyboardCommand.EnableA20Gate => false,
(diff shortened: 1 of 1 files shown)


src/Spice86.Core/Emulator/Devices/Input/Keyboard/Keyboard.cs
```diff
@@ -97,7 +97,7 @@ public override void WriteByte(ushort port, byte value) {
         switch (port) {
             case KeyboardPorts.Data:
                 _a20Gate.IsEnabled = Command switch {
-                    KeyboardCommand.SetOutputPort => (value & 2) > 0,
+                    KeyboardCommand.WriteOutputPort => (value & 2) > 0,
                     KeyboardCommand.EnableA20Gate => false,
                     KeyboardCommand.DisableA20Gate => true,
```
codengine: This should be
```
                   KeyboardCommand.EnableA20Gate => true,
                   KeyboardCommand.DisableA20Gate => false,
```

maximilien-noal: fixed.

maximilien-noal:
now rebased, but I still need to try and use the device scheduler instead.

PR #1413, opened 2025-10-23
Conditional memory breakpoint

kevinferrare:
### Description of Changes
Allows to define conditional breakpoints to watch for memory changes (experimental)


maximilien-noal (changes requested):
Seems like there is a regression:

edit: no it's not . Not really. It's the same on master.

Image

All memory views show the entire memory, not what the tab title says.

StartAddress says 0000:0000 instead of the start range.

Also, the name of the memory region (data segment, stack memory) isn't displayed anymore ?

PR #1356, opened 2025-09-17
Add support for saving the video card info to a file

xorangekiller:
The Spice86 builtin debugger is able to display a lot of information about the video card state, which is very useful for tracking down video-related issues. However prior to this PR, it wasn't possible to easily copy it and compare snapshots of the video card state at different points in time without manually copying each field or transcribing it in some other way, which was painful and time-consuming.

Therefore this PR adds a new context menu to the video debugger view that allows the user to save the whole video card state to disk as a JSON file. That makes it possible to run through a JSON formatting tool like `jq` and `diff` it much more easily.

Thanks to @maximilien-noal for getting me started by making the properties on the video card view read-only for me!

**Video Card Info debugger screen *before*:**

**Video Card Info debugger screen *after*:**

**New *Save Video Card Info* context menu:**

**New *Save Video Card Info* save file dialog:**

**Example video card info output file:**
```
$ cat VideoCardInfo.json | jq .
{
  "GeneralMiscellaneousOutputRegister": 99,
  "GeneralClockSelect": 0,
  "GeneralEnableRam": true,
  "GeneralVerticalSize": 400,
  "GeneralHorizontalSyncPolarity": 0,
  "GeneralVerticalSyncPolarity": 1,
  "GeneralIoAddressSelect": 1,
  "GeneralOddPageSelect": true,
  "GeneralInputStatusRegister0": 0,
  "GeneralCrtInterrupt": false,
  "GeneralSwitchSense": false,
  "GeneralInputStatusRegister1": 9,
  "GeneralDisplayDisabled": true,
  "GeneralVerticalRetrace": true,
  "DacReadIndex": 0,
  "DacWriteIndex": 0,
  "DacPixelMask": 255,
  "DacData": 0,
  "GraphicsDataRotate": 0,
  "GraphicsRotateCount": 0,
  "GraphicsFunctionSelect": 0,
  "GraphicsBitMask": 255,
  "GraphicsColorCompare": 0,
  "GraphicsReadMode": 0,
  "GraphicsWriteMode": 0,
  "GraphicsOddEven": false,
  "GraphicsShiftRegisterMode": 1,
  "GraphicsIn256ColorMode": false,
  "GraphicsMiscellaneousGraphics": 5,
  "GraphicsModeRegister": 0,
  "GraphicsGraphicsMode": true,
  "GraphicsChainOddMapsToEven": false,
  "GraphicsMemoryMap": 1,
  "GraphicsSetReset": 0,
  "GraphicsColorDontCare": 15,
[...]

PR #1422, opened 2025-10-24
fix: multi-exe games dumps shared the same folder

maximilien-noal:
### Description of Changes

This PR moves the default dumps folder from the current directory to the same directory where user breakpoints are saved:

into a subfolder of the current directory, named with the program's SHA256 hash.

### Rationale behind Changes

Games with multiple executables (COM and/or EXE files) would have their dumps overwrite each other. Not cool.

### Suggested Testing Steps

Already tested manually with Dune CD v3.7

image

### Breaking change

This is a breaking change, as the default dump dir is changed.


src/Spice86.Core/CLI/Configuration.cs
```diff
@@ -7,9 +7,17 @@ namespace Spice86.Core.CLI;
...
+using Spice86.Shared.Utils;
+
+using System.Security.Cryptography;
 
 /// <summary> Configuration for spice86, that is what to run and how. Set on startup. </summary>
 public sealed class Configuration {
+    /// <summary>
+    /// Gets the SHA 256 signature of the target <see cref="Exe"/>
+    /// </summary>
+    public string ProgramHash { get; internal set; } = "";
```
kevinferrare: It's a bit confusing to put the computed program hash here, this is the input config from command line. I would use another datastructure to comput it from dependecy injection maybe? And determine program dump file there, the pass it to the rrlevant classes

kevinferrare:
> I was wondering why some things used the current directory rather than the `SPICE86_DUMPS_FOLDER` when I added the video state save feature in #1356. LGTM.

I think the video state save feature should be uniformized with the rest of the dump/load features.
And also, currently, each dumper gets a folder in input at build time and they are built by the dependency management class which is probably not how this should be.

xorangekiller:
> > I was wondering why some things used the current directory rather than the `SPICE86_DUMPS_FOLDER` when I added the video state save feature in #1356. LGTM.
> 
> I think the video state save feature should be uniformized with the rest of the dump/load features. And also, currently, each dumper gets a folder in input at build time and they are built by the dependency management class which is probably not how this should be.

I was wondering whether the video state should be saved along with everything else. The reason I didn't is because it required a lot of plumbing through a bunch of different layers to do that (that was the approach I started with), and because I didn't think it was something useful to save in general. It's mostly helpful when you're specifically trying to track down video-related problems, in my opinion.

If you think it would be useful to dump by default along with everything else, I can revive my old stash from my first attempt at it after this PR is merged and raise another PR for it.

src/Spice86.Core/CLI/Configuration.cs
maximilien-noal: @copilot implement feedback on this PR, using a unit test driven approach.

See Configuration class, README.md and classes in the Dump namespace in the Spice86.Core folder for context.

First :

I think the video state save feature should be uniformized with the rest of the dump/load features.
And also, currently, each dumper gets a folder in input at build time and they are built by the dependency management class which is probably not how this should be.

second:

It's a bit confusing to put the computed program hash in Configuration, this is the input config from command line. I would use another datastructure to comput it from dependecy injection maybe? And determine program dump file there, the pass it to the relevant classes

context:


Description of Changes

This PR moves the default dumps folder from the current directory to the same directory where user breakpoints are saved:

into a subfolder of the current directory, named with the program's SHA256 hash.
Rationale behind Changes

Games with multiple executables (COM and/or EXE files) would have their dumps overwrite each other. Not cool.
Suggested Testing Steps

Already tested manually with Dune CD v3.7
image
Breaking change

This is a breaking change, as the default dump dir is changed.


The current approach of this PR that changes the CommandLineParser and the Configuration object is not appropriate.

maximilien-noal:
The irmprovments from this branch have been unit-tested, code has been reviewed, and a manual test with Dune reveals that it still works.

image

src/Spice86.Core/Emulator/Function/Dump/DumpFolderMetadata.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// <exception cref="FileNotFoundException">Thrown when the executable file doesn't exist.</exception>
+    public DumpContext(string? exePath, string? explicitDumpDirectory) {
+        ArgumentException.ThrowIfNullOrWhiteSpace(exePath);
+        
+        if (!File.Exists(exePath)) {
+            throw new FileNotFoundException($"Executable file not found: {exePath}", exePath);
+        }
+
+        ProgramHash = ComputeProgramHash(exePath);
+        DumpDirectory = DetermineDumpDirectory(explicitDumpDirectory);
```
codengine, line 43: Can you double check that there is no code path where this would throw a DirectoryNotExists exception wherever it is used?

maximilien-noal: Done.

maximilien-noal:
I will make the code take into account the old locations and move the files once.

PR #1427, opened 2025-10-26
New PIC, PIT, OPL3, Adlib Gold, DMA, many SB improvements

codengine:
### Description of Changes
New PIC, PIC, DMA, SIMD Instructions for audio processing, new OPL3, Adlib Gold and many, many other changes...

Currently missing or broken:

- Time Scaling
- Instruction-based limiting
- Breakpoint test is mostly broken (perhaps because of missing Instruction-based limiting) 

### Rationale behind Changes
I guess Max would be happy ;)

### Suggested Testing Steps
Test games thoroughly.

### DIFF
--- a/src/Spice86.Core/CLI/Configuration.cs
+++ b/src/Spice86.Core/CLI/Configuration.cs
@@ -7,6 +7,7 @@ using Spice86.Core.Emulator.Devices.Input.Mouse;
 using Spice86.Core.Emulator.Devices.Sound;
 using Spice86.Core.Emulator.Devices.Timer;
 using Spice86.Core.Emulator.Function;
+using Spice86.Core.Emulator.VM.CycleBudget;
 
 /// <summary> Configuration for spice86, that is what to run and how. Set on startup. </summary>
 public sealed class Configuration {
@@ -184,4 +185,7 @@ public sealed class Configuration {
 
     [Option(nameof(Xms), Default = null, Required = false, HelpText = "Enable XMS. Default is true.")]
     public bool? Xms { get; init; }
+
+    //TODO: Temporary fix, replace with real dependency injection
+    public ICyclesBudgeter? CyclesBudgeter { get; init; }
 }
--- a/src/Spice86.Core/Emulator/ProgramExecutor.cs
+++ b/src/Spice86.Core/Emulator/ProgramExecutor.cs
@@ -32,6 +32,7 @@ public sealed class ProgramExecutor : IDisposable {
     private readonly EmulatorBreakpointsManager _emulatorBreakpointsManager;
     private readonly EmulatorStateSerializer _emulatorStateSerializer;
     private readonly DumpFolderMetadata _dumpContext;
+    public event EventHandler? EmulationStopped;
 
     /// <summary>
     /// Initializes a new instance of <see cref="ProgramExecutor"/>
@@ -90,18 +91,26 @@ public sealed class ProgramExecutor : IDisposable {
     /// Starts the loaded program.
     /// </summary>
     public void Run() {
-        if (_loggerService.IsEnabled(LogEventLevel.Information)) {
-            _loggerService.Information("Starting the emulation loop");
-        }
-        if (_configuration.Debug) {
-            ToggleStartOrStopBreakpoint(BreakPointType.MACHINE_START, "Starting the emulated program in paused state.");
-            ToggleStartOrStopBreakpoint(BreakPointType.MACHINE_STOP, "Stopping the emulated program in paused state.");
-        }
-        _gdbServer?.StartServer();
-        _emulationLoop.Run();
+        try {
+            if (_loggerService.IsEnabled(LogEventLevel.Information)) {
+                _loggerService.Information("Starting the emulation loop");
+            }
 
-        if (_configuration.DumpDataOnExit is not false) {
-            _emulatorStateSerializer.SerializeEmulatorStateToDirectory(_dumpContext.DumpDirectory);
+            if (_configuration.Debug) {
+                ToggleStartOrStopBreakpoint(BreakPointType.MACHINE_START,
+                    "Starting the emulated program in paused state.");
+                ToggleStartOrStopBreakpoint(BreakPointType.MACHINE_STOP,
+                    "Stopping the emulated program in paused state.");
+            }
+
+            _gdbServer?.StartServer();
+            _emulationLoop.Run();
+
+            if (_configuration.DumpDataOnExit is not false) {
+                _emulatorStateSerializer.SerializeEmulatorStateToDirectory(_dumpContext.DumpDirectory);
+            }
+        } finally {
+            EmulationStopped?.Invoke(this, EventArgs.Empty);
         }
     }
 
--- a/src/Spice86.Core/Emulator/VM/CycleBudget/AdaptiveCyclesBudgeter.cs
+++ b/src/Spice86.Core/Emulator/VM/CycleBudget/AdaptiveCyclesBudgeter.cs
@@ -1,11 +1,11 @@
-﻿namespace Spice86.Core.Emulator.VM;
+﻿namespace Spice86.Core.Emulator.VM.CycleBudget;
 
 using Spice86.Core.Emulator.VM.CpuSpeedLimit;
 
 /// <summary>
 ///     Dynamically adjusts the per-slice CPU cycle budget so that emulation stays in sync with real time.
 /// </summary>
-internal sealed class AdaptiveCycleBudgeter {
+public sealed class AdaptiveCyclesBudgeter : ICyclesBudgeter {
     private const double MinAdaptiveScaleFactor = 0.25;
     private const double MaxAdaptiveScaleFactor = 4.0;
 
@@ -14,11 +14,11 @@ internal sealed class AdaptiveCycleBudgeter {
     private double _adaptiveCyclesPerSlice;
 
     /// <summary>
-    ///     Initializes a new instance of the <see cref="AdaptiveCycleBudgeter" /> class.
+    ///     Initializes a new instance of the <see cref="AdaptiveCyclesBudgeter" /> class.
     /// </summary>
     /// <param name="cyclesLimiter">Limiter exposing the user-configured cycles-per-millisecond target.</param>
     /// <param name="sliceDurationMilliseconds">Duration, in milliseconds, of each emulation slice.</param>
-    public AdaptiveCycleBudgeter(ICyclesLimiter cyclesLimiter, double sliceDurationMilliseconds) {
+    public AdaptiveCyclesBudgeter(ICyclesLimiter cyclesLimiter, double sliceDurationMilliseconds) {
         _cyclesLimiter = cyclesLimiter;
         _sliceDurationMilliseconds = Math.Max(0.001, sliceDurationMilliseconds);
         _adaptiveCyclesPerSlice = Math.Max(1, _cyclesLimiter.TargetCpuCyclesPerMs);
--- a/src/Spice86.Core/Emulator/VM/CycleBudget/ICyclesBudgeter.cs
+++ b/src/Spice86.Core/Emulator/VM/CycleBudget/ICyclesBudgeter.cs
@@ -1,0 +1,6 @@
+namespace Spice86.Core.Emulator.VM.CycleBudget;
+
+public interface ICyclesBudgeter {
+    int GetNextSliceBudget();
+    void UpdateBudget(double elapsedMilliseconds, long cyclesExecuted, bool cpuStateIsRunning);
+}
--- a/src/Spice86.Core/Emulator/VM/CycleBudget/StaticCyclesBudgeter.cs
+++ b/src/Spice86.Core/Emulator/VM/CycleBudget/StaticCyclesBudgeter.cs
@@ -1,0 +1,11 @@
+namespace Spice86.Core.Emulator.VM.CycleBudget;
+
+public class StaticCyclesBudgeter(int cycleBudget) : ICyclesBudgeter {
+    public int GetNextSliceBudget() {
+        return cycleBudget;
+    }
+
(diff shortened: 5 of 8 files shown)


maximilien-noal:
here is a test with the Adlib Gold and Dune:

https://www.youtube.com/watch?v=_cEEsm5jE7c

Dune is one of the very few games that can use not only the Adlib Gold, but also its AdLib Gold Surround Module, which is also emulated now.

Here is more information about the Adlib Gold card itself:

https://pcmidi.eu/goldlib.html

maximilien-noal:
- it has lots of unit tests
- the code is clean
- it has lots of documentation
- methods are public, enable ASM overrides of C# code to use it.
- the music of Dune are 100% correct
- Performance is good
- PCM doesn't make the GC crazy anymore
- 80% of code changes is in the sound stack.
- The PicEventQueue is a separate class from the PIC chip, but it's still a field of the PIC chip class and instantiated by it
- 

However, PCM still has dropouts and cutoffs. Tested with CFGCPU, in Debug mode, with fixed cycles at 3000.

Tested with commit 1b68ec00f36563a3fdd0cbb26d0ce1482db097b8

This at least has to be fixed.

codengine:
> PCM still has cutoffs in debug mode with cfgcpu

It should be much better now. I saw an IRQ storm in the logs because of cyclic Pump requests and very small chunks being read per pump.

One remark regarding the slowness: The time source for OPL and Adlib gold is the PIC Ticker. If the emulation runs much slower than the target cycles, it will be played back too slowly.

Another thing is an open to do that I left in code: Time Scaling is unimplemented right now, the property is there but has no effect.

codengine:
Another issue is that the Breakpoint Tests won't work reliably and will lockup my computer... I have no clue yet what the cause is.

src/Spice86.Core/Emulator/IOPorts/IoSystem.cs
```diff
@@ -0,0 +1,234 @@
...
+
+using Spice86.Core.Emulator.CPU;
+using Spice86.Core.Emulator.Devices;
+using Spice86.Shared.Interfaces;
+
+/// <summary>
+///     Coordinates delegate-based registration for emulated I/O port reads and writes on top of the shared
+///     <see cref="IOPortDispatcher" />.
+/// </summary>
+public sealed class IoSystem(
```
maximilien-noal, line 11: This should be named IOPortDispatcher :)

src/Spice86.Core/Emulator/IOPorts/IoSystem.cs
```diff
@@ -0,0 +1,234 @@
...
+            handler.WriteByte(port, value);
+        }
+    }
+
+    private DelegatePortHandler GetOrCreateHandler(ushort port) {
+        if (_handlers.TryGetValue(port, out DelegatePortHandler? existing)) {
+            return existing;
+        }
+
+        var handler = new DelegatePortHandler(this, state, failOnUnhandledPort, loggerService);
```
maximilien-noal, line 158: What if we want to fail on unhandled port with -f option ?

It helps when discovering why a game doesn't work, or when a new sound device is being implemented (like the Gravis Ultrasound)

maximilien-noal (changes requested):
No issues with the performance, code quality, code clarity, or the sound effects or music.

Just two requests for changes.

maximilien-noal:
Breakpoint test is mostly broken (perhaps because of missing Instruction-based limiting) needs to be fixed before merging.

The license of the emulator needs to be compatible too.

maximilien-noal:
GDB and overrides with Cryogenic and BaK need to be tested (by me or any other person) before this is merged.

src/Spice86.Core/Emulator/Devices/Timer/WallClock.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    /// <summary>
+    ///     Gets the singleton <see cref="SystemTimeProvider" /> instance backed by the system clock.
+    /// </summary>
+    public static SystemTimeProvider Instance { get; } = new();
+
+    /// <summary>
+    ///     Gets the current UTC timestamp as reported by <see cref="DateTime.UtcNow" />.
+    /// </summary>
+    public DateTime UtcNow => DateTime.UtcNow;
```
maximilien-noal, line 11: UtcNow and Now are super slow - can this use Stopwatch.GetTimestamp ?

codengine: How did you measure that its slow?

maximilien-noal: Environment.TickCount64 gets a long value representing the number of ticks since the host system started, but its super slow too.

maximilien-noal:
I created a branch out of this branch in order to eventually fix all the things mentionned.

src/Spice86.Core/Emulator/Devices/Timer/WallClock.cs
maximilien-noal: It's slow (for an emulator) because it doesn't use high precision performance counters.

It can be measured in a dotnet benchmark.

It was measured to be slow at the start of the project.

Only the Stopwatch uses high precision performance counters.

Even then, both dosbox (in dosbox.cpp) and the counterActivators on the master branch try not to call it too often in tight loops (such as the emulation loop).

src/Spice86.Core/Emulator/CPU/CPU.cs
```diff
@@ -71,6 +72,7 @@ public class Cpu : IInstructionExecutor, IFunctionHandlerProvider {
     private readonly CircularBuffer<SegmentedAddress> _lastAddresses = new(20);
 
     private readonly IOPortDispatcher _ioPortDispatcher;
+    private readonly PicPitCpuState _picPitCpuState;
```
kevinferrare, line 75: Move that logic to emulation loop.
Same for cfgcpu.

src/Spice86.Core/Emulator/Devices/ExternalInput/PicEventQueue.cs
```diff
@@ -0,0 +1,325 @@
...
+using Spice86.Core.Emulator.VM;
+using Spice86.Shared.Interfaces;
+
+/// <summary>
+///     Manages deterministic scheduling of PIC events relative to the CPU tick index.
+/// </summary>
+/// <remarks>
+///     Pools entries to avoid allocations and relies on the shared <see cref="PicPitCpuState" /> for cycle accounting.
+/// </remarks>
+internal sealed class PicEventQueue {
```
kevinferrare, line 14: Not really linked to Pic? Move it with rename to a new package under VM/EmulationLoop?
I would rename that to EmulationLoopEventQueue

Same remark for the PicEntry or PicPitCpuState

src/Spice86.Core/Emulator/Devices/ExternalInput/PicEventQueue.cs
```diff
@@ -0,0 +1,325 @@
...
+        Initialize();
+    }
+
+    /// <summary>
+    ///     Resets the queue and returns all entries to the free list.
+    /// </summary>
+    /// <remarks>
+    ///     Reinitialises the pool to its pristine state, mirroring the bootstrap performed during construction.
+    /// </remarks>
+    public void Initialize() {
```
kevinferrare, line 43: Do we need to pre-allocate events? Wouldn't code be simpler if we just allocated them on the go?
Wouldnt something like that work?
https://learn.microsoft.com/fr-fr/dotnet/api/system.collections.generic.priorityqueue-2?view=net-8.0

src/Spice86.Core/Emulator/Devices/ExternalInput/PicEventQueue.cs
```diff
@@ -0,0 +1,325 @@
...
+/// </remarks>
+internal sealed class PicEventQueue {
+    private const int PicQueueSize = 8192; // Larger value from DosBox-X. Staging uses 512.
+    private readonly PicPitCpuState _cpuState;
+    private readonly PicEntry[] _entryPool = new PicEntry[PicQueueSize];
+
+    private readonly ILoggerService _logger;
+
+    private PicEntry? _freeEntry;
+    private bool _inEventService;
```
kevinferrare, line 22: Do we need _inEventService and _srvLag? Why not doing multithread sync with a standard lock?

src/Spice86.Core/Emulator/Devices/ExternalInput/PicEventQueue.cs
```diff
@@ -0,0 +1,325 @@
...
+    ///     after draining ready handlers.
+    /// </remarks>
+    public bool RunQueue() {
+        _cpuState.CyclesLeft += _cpuState.Cycles;
+        _cpuState.Cycles = 0;
+        if (_cpuState.CyclesLeft <= 0) {
+            return false;
+        }
+
+        double indexNd = _cpuState.TickIndexNd;
```
kevinferrare, line 198: TickIndexNd => TickIndexInCycles?

This whole thing is very complex.
Time is measured in 3 ways as far as I understand:
- Miliseconds
- Cycles
- Index

Why not using a single unit for the events which would be "time in nanoseconds"?
In C# there is System.Diagnostics.Stopwatch.GetTimestamp (used in current implementation)

We could have a clock that gets the current time in nanoseconds from the cpu cycles / a periodic Stopwatch call to avoid having bad perfs and just call that here. This would separate the concerns of a time based priority queue and time counting.

src/Spice86.Core/Emulator/Devices/ExternalInput/Intel8259Pic.cs
```diff
@@ -0,0 +1,459 @@
...
+
+    /// <summary>
+    ///     Determines whether the candidate IRQ outranks the currently active IRQ based on the rotation state.
+    /// </summary>
+    /// <param name="candidate">IRQ line to test.</param>
+    /// <param name="current">Currently active IRQ line, or 8 when idle.</param>
+    /// <returns><see langword="true" /> when the candidate has a higher priority; otherwise <see langword="false" />.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private bool HasHigherPriority(byte candidate, byte current) {
+        if (current == 8) {
```
kevinferrare, line 126: 8 => Use NoPendingIrq? (move it from DualPic to here maybe?)

src/Spice86.Core/Emulator/Devices/ExternalInput/Intel8259Pic.cs
```diff
@@ -0,0 +1,459 @@
...
+        }
+    }
+
+    /// <summary>
+    ///     Validates that an IRQ number is within the supported 0-7 range.
+    /// </summary>
+    /// <param name="value">IRQ value to validate.</param>
+    /// <param name="caller">The name of the calling member populated automatically.</param>
+    /// <returns><see langword="true" /> when the value is valid; otherwise <see langword="false" />.</returns>
+    private bool ValidateIrq(byte value, [CallerMemberName] string? caller = null) {
```
kevinferrare, line 241: Are there valid cases where value is >7 and we dont want the emulator to crash because something very wrong happenened?

src/Spice86.Core/Emulator/Devices/ExternalInput/Intel8259Pic.cs
```diff
@@ -0,0 +1,459 @@
...
+    /// <summary>
+    ///     Transfers a pending IRQ into service without directly invoking the CPU.
+    /// </summary>
+    /// <param name="value">IRQ line number.</param>
+    public void StartIrq(byte value) {
+        if (!ValidateIrq(value)) {
+            return;
+        }
+
+        InterruptRequestRegister &= (byte)~(1 << value);
```
kevinferrare, line 346: This is very confusing. RaiseIrq is the method which should update this.

Shouldnt logic be:
- device wants to raise IRQ
- event is added to the event queue
- event is processed, calling RaiseIrq
- later, CPU checks whether there is a pending IRQ or not via ComputeVectorNumber

src/Spice86.Core/Emulator/Devices/ExternalInput/DualPic.cs
```diff
@@ -1,207 +1,900 @@
...
+        if (selectedIrq == NoPendingIrq) {
+            _logger.Error("PIC {Controller}: IRQ 2 is active, but IRQ is not active on the secondary controller.",
+                nameof(PicController.Secondary));
+
+            return;
+        }
+
+        _secondaryPic.StartIrq(selectedIrq);
+        _primaryPic.StartIrq(2);
+        _cpuState.CpuHwInterrupt((byte)(_secondaryPic.InterruptVectorBase + selectedIrq));
```
kevinferrare, line 636: Why setting this to a separate state object since this is going to be returned by the caller (ComputeVectorNumber)
This is also called from RunQueue but this should not do anything more than running the event queue and setup the PIC if an event for that is there

src/Spice86.Core/Emulator/Devices/ExternalInput/Intel8259Pic.cs
```diff
@@ -0,0 +1,459 @@
...
+        // Those bits have an impact on whether the cpu emulation should be paused or not.
+        if ((InterruptRequestRegister & change & InServiceRegisterInverted) != 0) {
+            CheckForIrq();
+        }
+    }
+
+    /// <summary>
+    ///     Re-evaluates pending IRQs after an end-of-interrupt is issued.
+    /// </summary>
+    public void CheckAfterEoi() {
```
kevinferrare, line 208: You dont need that if you simplify the design meaning that IRQs are checked by the CPU when it needs them and not at other places.

maximilien-noal:
Let's merge this. :)

maximilien-noal (approved):

maximilien-noal:
performance is great.

only tests need to pass.

PR #1428, opened 2025-10-26
Respect the VGA Vertical Sync End register's Write Protect bit

### LINKED ISSUE #1390: EXODUS displays half a screen; Monkey Island and Tom Long: menu/overlay doesn't show
for this issue:

https://github.com/barotto/VGATEST

https://www.pcjs.org/software/pcx86/test/vga/blackbook/

xorangekiller:
According to the VGA spec, the Vertical Sync End register in the CRT Controller is supposed to prevent CRT Controller registers 0 - 7 from being written. This bit was unused in the earlier EGA spec. This bit is set by default in DOS INT10 video modes 0x00 - 0x0E that were introduced prior to VGA to enable interoperability of programs that use CGA, EGA, and early text modes to work with VGA video cards, such as the one emulated by Spice86.

Prior to this, we already had the correct definition of the `WriteProtect` bit in the `VerticalSyncEndRegister`, but it wasn't being used anywhere. It was simply ignored. This worked correctly for games that used VGA compatible video modes, and for some games that used earlier EGA compatible video modes but didn't set any of the reserved bits that were later used by VGA, but not for all earlier games.

To fix it, I made the CRT Controller properly respect the Vertical Sync End register's Write Protect bit, which in turn enables compatibility with more games that use EGA (and possibly CGA - although that remains untested). I also added more detail to some of the comments and log messages that I used to debug this problem, which will hopefully help anyone looking at it in the future to pick it up more quickly.

I tested it with Exodus, which now works correctly, and regression tested it with Dune. Everything seems to work correctly.
image
image

This fixes #1390.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ Also, while in Seer, set Settings/Configuration/Assembly/Disassembly Mode to
 |  | HMA is implemented |
 |  | No paging support |
 | **Graphics** | Text modes, VGA, EGA, and CGA implemented |
+|  | EGA and CGA modes are best effort (you may find bugs) |
 |  | No VESA support |
 | **DOS** | Partial int 21h implementation (DOS 5.0) |
 | **Input** | Keyboard and mouse supported |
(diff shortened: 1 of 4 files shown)


codengine:
Awesome job!

PR #1448, opened 2025-10-29
add ems unit tests

maximilien-noal:
This PR adds extensive unit and integration tests to EMS.


maximilien-noal:
@copilot Make the rest of the EMS integration test program pass.

I know for a fact that the implementation is correct - only a few remaining EMS integration tests programs written in ASM use EMS incorrectly.

codengine:
@maximilien-noal There is just that one unused import that Copilot complained about.

PR #1457, opened 2025-11-01
PR fix: passing tests for the New PIT, PIC, NukedOPL, new Sound Blaster, and Adlib Gold PR

maximilien-noal:
### Description of Changes

PR that intents to make all the tests pass on the PIT/PIC/Device event scheduling / budget-based cycles limiting / proper SB, OPL, and Adlib Gold emulation PR.

### Rationale behind Changes

Ensure that breakpoints, ASM overrides, and such still work by fixing unit tests.

### Suggested Testing Steps

None, this is only a step towards fully fixing the target PR.

### DIFF
--- a/src/Spice86.Core/Emulator/Devices/Timer/PitTimer.cs
+++ b/src/Spice86.Core/Emulator/Devices/Timer/PitTimer.cs
@@ -298,7 +298,7 @@ public sealed class PitTimer : IDisposable, IPitControl, ITimeMultiplier {
         };
     }
 
-    private void UpdateChannelDelay(ref PitChannel channel) {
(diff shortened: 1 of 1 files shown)


maximilien-noal:
@copilot GDB support should have unit tests and integration tests.

Integration tests means using a GDB client to talk and excercise every function of the server, in headless mode (without even the headless Avalonia mode), and in UI enabled mode.

Problem is, there is no (to my knowledge) easily integrable .NET GDB client, not without GDB itself anyway.

The ideal solution would be to depend on the GDB client mentionned in the README;md and on GDB itself, and drive them all with integration tests.

Don't forget the unit tests.

Reason: 

InstructionsPerSecond mode is important for GDB users

GDB support had regressions several times in the past.

PR #1467, opened 2025-11-03
Cfgcpu fixes

codengine:
### Description of Changes
Many new instructions and bugfixes for already existing instructions.

### DIFF
--- a/src/Spice86.Core/Emulator/CPU/CfgCpu/ParsedInstruction/Instructions/Mixins/Bound.mixin
+++ b/src/Spice86.Core/Emulator/CPU/CfgCpu/ParsedInstruction/Instructions/Mixins/Bound.mixin
@@ -53,7 +53,7 @@ public partial class {{ moxy.Class.Name }} : InstructionWithModRm {
 
         ValueNode offset = builder.ModRm.MemoryOffsetToNode(ModRmContext);
         ValueNode increment = builder.Constant.ToNode({{ if Size == 16 }}(ushort){{ElementSize}}{{ else }}(uint){{ElementSize}}{{ end }});
-        ValueNode upperOffset = new BinaryOperationNode(builder.AddressType(this), offset, Operation.PLUS, increment);
+        ValueNode upperOffset = new BinaryOperationNode(builder.AddressType(this), offset, BinaryOperation.PLUS, increment);
         ValueNode upperPointer = builder.Pointer.ToSegmentedPointer(builder.SType({{Size}}), (SegmentRegisterIndex)ModRmContext.SegmentIndex.Value, upperOffset);
 
         return new InstructionNode(


src/Spice86.Core/Emulator/CPU/Flags.cs
```diff
@@ -11,10 +11,12 @@ public class Flags {
...
         [CpuModel.ZET_86] = new(bitsAlwaysOn: [1], bitsAlwaysOff: [3, 5, 15]),
         // Real CPUs
+        // 8086: high bits 12..15 are always set, 3 and 5 are undefined and treated as 0
         [CpuModel.INTEL_8086] = new(bitsAlwaysOn: [1, 12, 13, 14, 15], bitsAlwaysOff: [3, 5]),
-        // Since we dont handle IO privilege (12 / 13) and nested task (14), let's put them as always off
-        [CpuModel.INTEL_80286] = new(bitsAlwaysOn: [1], bitsAlwaysOff: [3, 5, 12, 13, 14, 15]),
-        [CpuModel.INTEL_80386] = new(bitsAlwaysOn: [1], bitsAlwaysOff: [3, 5, 12, 13, 14, 15])
+        // 80286/80386 in real-mode: allow IOPL (12/13) and NT (14) to be set by POPF(D), like DOSBox.
+        // Keep only undefined bits 3 and 5 always off, and bit 1 always on.
+        [CpuModel.INTEL_80286] = new(bitsAlwaysOn: [1], bitsAlwaysOff: [3, 5, 15]),
```
kevinferrare: Let's keep this as it is now, flags 13 and 14 are not updatable I think according to single step tests

src/Spice86.Core/Emulator/CPU/CfgCpu/ParsedInstruction/Instructions/Mixins/Enter.mixin
```diff
@@ -24,20 +24,32 @@ public partial class {{ moxy.Class.Name }} : EnterInstruction {
...
         byte level = helper.InstructionFieldValueRetriever.GetFieldValue(LevelField);
-        helper.Stack.Push{{Size}}(helper.State.{{BasePointerRegister}});
-        level &= 0x1f;
-        {{StackPointerType}} framePtr = helper.State.{{StackPointerRegister}};
-        for (int i = 0; i < level; i++) {
-            helper.State.{{BasePointerRegister}} -= {{Size}}/8;
-            helper.Stack.Push{{Size}}(helper.State.{{BasePointerRegister}});
+        level &= 0x1F;
+        {{StackPointerType}} pointerSize = ({{StackPointerType}})({{Size}} / 8);
+        {{StackPointerType}} oldBasePointer = helper.State.{{BasePointerRegister}};
```
kevinferrare, line 29: old -> initial?

src/Spice86.Core/Emulator/CPU/CfgCpu/ParsedInstruction/Instructions/Mixins/Enter.mixin
```diff
@@ -24,20 +24,32 @@ public partial class {{ moxy.Class.Name }} : EnterInstruction {
     public override void Execute(InstructionExecutionHelper helper) {
         ushort storage = helper.InstructionFieldValueRetriever.GetFieldValue(StorageField);
         byte level = helper.InstructionFieldValueRetriever.GetFieldValue(LevelField);
-        helper.Stack.Push{{Size}}(helper.State.{{BasePointerRegister}});
-        level &= 0x1f;
-        {{StackPointerType}} framePtr = helper.State.{{StackPointerRegister}};
-        for (int i = 0; i < level; i++) {
-            helper.State.{{BasePointerRegister}} -= {{Size}}/8;
-            helper.Stack.Push{{Size}}(helper.State.{{BasePointerRegister}});
+        level &= 0x1F;
```
kevinferrare, line 27: Could you ask codex what it fixed exactly? Algorithm is difficult to read, it converted all the pushes to stack arithmetic

src/Spice86.Core/Emulator/CPU/CfgCpu/ParsedInstruction/Instructions/Cpuid.cs
```diff
@@ -0,0 +1,22 @@
...
+using Spice86.Core.Emulator.CPU.CfgCpu.Ast.Builder;
+using Spice86.Core.Emulator.CPU.CfgCpu.Ast.Instruction;
+using Spice86.Core.Emulator.CPU.CfgCpu.InstructionExecutor;
+using Spice86.Core.Emulator.CPU.CfgCpu.ParsedInstruction.Prefix;
+using Spice86.Core.Emulator.CPU.Exceptions;
+using Spice86.Shared.Emulator.Memory;
+using Spice86.Shared.Utils;
+
+public class Cpuid(SegmentedAddress address, InstructionField<ushort> opcodeField, List<InstructionPrefix> prefixes)
+    : CfgInstruction(address, opcodeField, prefixes, 1) {
```
kevinferrare, line 11: Why not just throwing invalid opcode exception in the parser?
If you really want to do it here, calling HandleCpuException here is not correct I think, it's better to rely on the centralized cfgcpu handler.

src/Spice86.Core/Emulator/CPU/Alu16.cs
```diff
@@ -101,17 +99,18 @@ public override ushort Rcl(ushort value, byte count) {
             return value;
         }
 
+        bool oldCarry = _state.CarryFlag;
         int carry = value >> 16 - count & 0x1;
         ushort res = (ushort)(value << count);
         int mask = (1 << count - 1) - 1;
         res = (ushort)(res | (value >> 17 - count & mask));
-        if (_state.CarryFlag) {
```
kevinferrare, line 108: This doesnt fix anything I think. oldCarry is never used after CarryFlag is updated

src/Spice86.Core/Emulator/CPU/Alu16.cs
```diff
@@ -101,17 +99,18 @@ public override ushort Rcl(ushort value, byte count) {
...
+        if (oldCarry) {
             res = (ushort)(res | 1 << count - 1);
         }
 
         _state.CarryFlag = carry != 0;
         bool msb = (res & MsbMask) != 0;
-        _state.OverflowFlag = msb ^ _state.CarryFlag;
+        _state.OverflowFlag = _state.CarryFlag ^ msb;
```
kevinferrare, line 113: isnt xor bijective? what does this fixes?

src/Spice86.Core/Emulator/CPU/Alu16.cs
```diff
@@ -145,7 +145,8 @@ public override ushort Rol(ushort value, byte count) {
         res = (ushort)(res | value >> 16 - count);
         _state.CarryFlag = carry != 0;
         bool msb = (res & MsbMask) != 0;
-        _state.OverflowFlag = msb ^ _state.CarryFlag;
+        bool lsb = (res & 0x01) != 0;
```
kevinferrare, line 148: Could you ask codex what it attempted to fix here? I am wondering if this is correct or not since this causes machine test results to change. Same for ror

src/Spice86.Core/Emulator/CPU/Alu16.cs
```diff
@@ -246,11 +245,11 @@ public override ushort Shr(ushort value, int count) {
             return value;
         }
 
-        ushort msb = (ushort)(value & MsbMask);
-        _state.OverflowFlag = msb != 0;
+        bool msb = (value & MsbMask) != 0;
         SetCarryFlagForRightShifts(value, count);
         ushort res = (ushort)(value >> count);
         UpdateFlags(res);
+        _state.OverflowFlag = count == 1 && msb;
```
kevinferrare, line 252: Nice catch :)

src/Spice86.Core/Emulator/CPU/Alu16.cs
```diff
@@ -79,9 +79,7 @@ public override uint Mul(ushort value1, ushort value2) {
         bool upperHalfNonZero = (res & 0xFFFF0000) != 0;
         _state.OverflowFlag = upperHalfNonZero;
         _state.CarryFlag = upperHalfNonZero;
-        SetZeroFlag(res);
-        SetParityFlag(res);
-        SetSignFlag((ushort)res);
+        _state.ZeroFlag = (res & 0xFFFF) == 0;
```
kevinferrare, line 82: Changed code is undefined behaviour. Are you sure about that?

src/Spice86.Core/Emulator/CPU/Flags.cs
codengine: Actual CPUs (80286/80386 real mode). Intel’s manuals state that POPF/IRET may update IOPL when CPL=0 (which is the only mode we run) and that NT (bit 14) is also writable. Only undefined bits 3 and 5 are forced to zero, and bit 1 is always set. Masking off 12/13/14 makes POPF behave unlike real hardware even in real mode.

DosBox, Bochs, 86Box. Each one mirrors that behavior: they mask IOPL/IF based on CPL but do not permanently clear bits 12–14 (see dosbox-staging/src/cpu/cpu.cpp:2108-2200, bochs/cpu/flag_ctrl.cc:209-335, 86Box/src/cpu/x86_ops_flag.h:186-210). Our emulation should stay aligned with them.

src/Spice86.Core/Emulator/CPU/CfgCpu/ParsedInstruction/Instructions/Mixins/Enter.mixin
codengine: • The ENTER sequence has three required phases per Intel’s spec:

  1. Push the current BP/EBP.
  2. If level > 0, copy the caller’s frame pointers: for i = 1 .. level-1 push the word/dword at [old BP - i*ptrSize], then push the temporary frame pointer.
  3. Set BP/EBP to that frame pointer and subtract the local storage size from SP/ESP.

  Our old implementation didn’t do that:

  - After the initial push, it mutated BP/EBP inside the loop (helper.State.{BasePointerRegister} -= Size/8) and pushed the raw register value. That pushes decrementing addresses, not the contents of the previous frame. So any nesting
    level > 0 produced a bogus chain of pointers, breaking debuggers/exception unwinding and code that walks stack frames.
  - We never pushed the temporary frame pointer value when level != 0, so the last pointer in the chain was missing entirely.
  - We moved the stack pointer by simply subtracting the storage count, which was correct for 16-bit BP but wrong for 32-bit (no wrapping type, no separate pointer-sized arithmetic).

  The new code fixes all of that:

  - It snapshots the original BP/EBP and SP/ESP, manually writes the saved pointers through helper.Memory.UInt{Size} so we can copy the values sitting in the caller’s frame (just like the spec), and only updates BP/EBP once after the
    chain is built.
  - It pushes the temporary frame pointer after mirroring the requested number of ancestors.
  - It adjusts SP/ESP using the appropriate pointer type, so 32-bit ENTER obeys proper wrapping semantics.

  In short, the previous implementation produced malformed frame lists whenever level > 0; the new logic now mirrors what real CPUs (and DosBox/Bochs/86Box) do, so stack unwinding and debuggers see the expected layout.

src/Spice86.Core/Emulator/CPU/Alu16.cs
codengine: Its a stylistic change, to keep it aligned with the other instructions.

src/Spice86.Core/Emulator/CPU/Alu16.cs
codengine: Yes it is. It was done to match the documentation and improve readability.

src/Spice86.Core/Emulator/CPU/Alu16.cs
codengine: For ROL we didn’t change the actual behavior—we just rewrote the overflow assignment in a way that makes the Intel rule obvious.

  - In the current code (src/Spice86.Core/Emulator/CPU/Alu16.cs:96-114) we compute the rotate result res, set CarryFlag to the bit that dropped off the left end, then set OverflowFlag = msb ^ lsb, where msb and lsb are the new bit 15 and
    bit 0 of the result.
  - In the pre-change code (git show 79a436b:src/Spice86.Core/Emulator/CPU/Alu16.cs, same region) we had _state.OverflowFlag = msb ^ _state.CarryFlag.

  Because a rotate is just a series of single-bit rotates, the bit that lands in CF is always the same bit that lands in bit 0 of the result. (Intel’s spec describes ROL as copying the high-order bit into CF and into bit 0; see also
  DosBox’s macro in dosbox-staging/src/cpu/instructions.h:228-238, which sets CF = lf_resw & 1.) Bochs implements the same rule in bochs/cpu/shift16.cc:368-418. So msb ^ lsb and msb ^ _state.CarryFlag are identical expressions; we simply
  rewrote it to mirror the spec (“OF = MSB xor LSB”) and to match the 8‑/32‑bit helpers stylistically.

src/Spice86.Core/Emulator/CPU/Alu16.cs
codengine: In the previous implementation Mul(ushort, ushort) called SetZeroFlag, SetParityFlag, and SetSignFlag on the full 32‑bit product, so the low 16 bits dragged PF/SF/ZF around like it was a normal logic op. That doesn’t match real
  hardware: unsigned MUL defines CF/OF based on the high half of the product, while the other flags are either undefined (per Intel) or left as the low-half result on real CPUs/emulators.

  When we compared against the reference emulators:

  - DosBox (src/cpu/instructions.h:576-607) sets ZF to true when the low 8/16/32 bits of the product are zero, but it doesn’t update PF/SF.
  - Bochs (bochs/cpu/mult32.cc:27-45) calls SET_FLAGS_OSZAPC_LOGIC_* on the low half, so ZF reflects that lower word.
  - 86Box (src/cpu/808x.c:3668-3719) calls set_pzs after MUL, which sets ZF/PF/SF based on the stored result.

  So we trimmed the old “set three flags” sequence down to just ZF, and we compute it from the same lower half those emulators use: _state.ZeroFlag = (res & 0xFFFF) == 0;. That keeps CF/OF handling unchanged but aligns our ZF with the
  references and with the updated machine tests.

src/Spice86.Core/Emulator/CPU/CfgCpu/ParsedInstruction/Instructions/Cpuid.cs
codengine: Mind to propose a proper implementation of this case, as discussed in Discord? (with CPUID being used as a probe for the CPU type)

kevinferrare: I think just throwing a new CpuInvalidOpcodeException would work here, will be caught by the handler and cfgcpu will call HandleCpuException like it does for other cpu faults.

codengine:
> I don't know about the CPU, but I've got this build error:
> 
> ```
> Spice86.Core net8.0 a échoué avec 2 erreur(s) (22,9s)
>   C:\Users\noalm\source\repos\Spice86\src\Spice86.Core\obj\Debug\net8.0\generated\Morris.Moxy\Morris.Moxy.RoslynIncrementalGenerator\Spice86.Core.Emulator.CPU.CfgCpu.ParsedInstruction.Instructions.Bound32.Bound.1.Moxy.g.cs(51,92): error CS0103: Le nom 'Operation' n'existe pas dans le contexte actuel
>   C:\Users\noalm\source\repos\Spice86\src\Spice86.Core\obj\Debug\net8.0\generated\Morris.Moxy\Morris.Moxy.RoslynIncrementalGenerator\Spice86.Core.Emulator.CPU.CfgCpu.ParsedInstruction.Instructions.Bound16.Bound.1.Moxy.g.cs(51,92): error CS0103: Le nom 'Operation' n'existe pas dans le contexte actuel
> ```
> 
> same as CI/CD error...?

Yes, and I dont know why. The build works locally for me

PR #1489, opened 2025-11-10
Improve performance measurements

codengine:
Two things are changed here:
- The previous implementation does not reflect rapid execution speed changes due to the rolling average
- TickCount64 followed by the division on each tick is expensive

The values for the sampling can be discussed but overall, the performance measurement now reflects the "current" performance better and cheap checks guard costly calculations

### DIFF
--- a/src/Spice86.Core/Emulator/VM/EmulationLoop.cs
+++ b/src/Spice86.Core/Emulator/VM/EmulationLoop.cs
@@ -1,5 +1,7 @@
 ﻿namespace Spice86.Core.Emulator.VM;
 
+using Serilog.Events;
+
 using Spice86.Core.Emulator.CPU;
 using Spice86.Core.Emulator.Devices.DirectMemoryAccess;
 using Spice86.Core.Emulator.Devices.Timer;
@@ -12,7 +14,6 @@ using Spice86.Shared.Emulator.Memory;
 using Spice86.Shared.Interfaces;
 
 using System.Diagnostics;
-
 
 /// <summary>
 /// This class orchestrates the execution of the emulated CPU, <br/>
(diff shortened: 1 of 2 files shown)


src/Spice86.Shared/Diagnostics/PerformanceMeasurer.cs
```diff
@@ -26,45 +25,57 @@ public class PerformanceMeasurer : IPerformanceMeasureReader, IPerformanceMeasur
...
 
     private bool IsLastMeasurementExpired(long newTimeInMilliseconds) =>
         newTimeInMilliseconds - _firstMeasureTimeInMilliseconds > WindowSizeInSeconds * 1000;
 
-    private static long ApproxRollingAverage(long measureAverage, long valuePerSecond, long sampledMetricsCount) {
-        measureAverage -= measureAverage / Math.Max(sampledMetricsCount, 1);
-        measureAverage += valuePerSecond / Math.Max(sampledMetricsCount, 1);
-        return measureAverage;
+    private static long SmoothingAverage(long currentAverage, long latestValuePerSecond) {
+        const double alpha = 0.2;
```
maximilien-noal, line 78: What is alpha ? Why this value ?

maximilien-noal (approved):

src/Spice86.Core/Emulator/VM/EmulationLoop.cs
```diff
@@ -122,14 +129,34 @@ private void RunOnce() {
...
         _pauseHandler.WaitIfPaused();
         _cpu.ExecuteNext();
-        _performanceMeasurer.UpdateValue(_cpuState.Cycles);
+        MaybeUpdatePerformanceMeasurements();
         _timer.Tick();
         _dmaController.PerformDmaTransfers();
         _cyclesLimiter.RegulateCycles(_cpuState);
     }
 
+    private void MaybeUpdatePerformanceMeasurements() {
```
kevinferrare, line 138: This should have been done in _performanceMeasurer. Also this should not introduce a new counter but rely on cycles to know when to activate.

PR #1490, opened 2025-11-11
Fix: corruption in Dune

codengine:
The issue was caused by misdetecting an SMC of a loop-instruction


kevinferrare (approved):

kevinferrare:
Quite happy this is fixed, thank you!

PR #1527, opened 2025-11-17
Feature: PS/2 micocrontroller, full INT9H, fixed Avalonia keyboard enum usage, fix UI/Emu input thread race, full keyboard command and code sets

maximilien-noal:
### Description of Changes

Implements the full PS/2 micro controller (INTEL 8049 chip), the internal keyboard buffer, the proper UI VS Emu thread concurrent input queue access handling, typematic events, keyboard delay event, keyboard reset event, code sets, and IBM AT vs XT code set conversions, and keyboard key interception (can be overriden by TSRs), and all the keyboard commands.

### Rationale behind Changes

Implements most of https://github.com/OpenRakis/Spice86/issues/329

Previously, the PS/2 micro controller, and full INT9H were partially or entirely unimplemented. Same for the rest.

It should vastly improve the usability and compatibility of the keyboard and PS/2 micro controller emulation.

It already makes the keyboard easier to use, but there is still a big DOS issue with a key IOCTL INT21H function to fix for the DOS console.

The reason why essentially is that the keyboard event flow now properly flows from: UI → InputEventQueue → EmulationLoop.ProcessAllPendingInputEvents() → PS2Keyboard → BIOS keyboard buffer.

Since this is focused on hardware, I left DOS out of this PR.

### Suggested Testing Steps

Already tested with:

Dune
Dune 2
Jill of the Jungle
Krondor
Hocus Pocus
Another World

Compared to master, it's especially better with Jill of the Jungle.

### Skipped

DOS Console issues...
DOS Codepage issue...
Mouse AUX and mouse notifications, and mouse frame...
Print Screen special support...
Pause key special support...

### DIFF
--- a/src/Spice86.Core/Emulator/Devices/Input/Keyboard/Intel8042Controller.cs
+++ b/src/Spice86.Core/Emulator/Devices/Input/Keyboard/Intel8042Controller.cs
@@ -11,7 +11,6 @@ using Spice86.Shared.Interfaces;
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 
 /// <summary>
 /// Emulates the Intel 8042 keyboard controller, providing PS/2 keyboard and mouse (auxiliary device) port functionality
@@ -30,9 +29,7 @@ public partial class Intel8042Controller : DefaultIOPortHandler {
     /// </summary>
     private const double PortDelayMs = 0.300; // 0.3ms
 
-    private readonly BufferEntry[] _buffer = new BufferEntry[BufferSize];
-    private int _bufferStartIdx = 0;
-    private int _bufferNumUsed = 0;
+    private readonly Queue<BufferEntry> _buffer = new(BufferSize);
 
     private int _waitingBytesFromAux = 0;
     private int _waitingBytesFromKbd = 0;
@@ -250,8 +247,7 @@ public partial class Intel8042Controller : DefaultIOPortHandler {
         _status.IsDataPending = false;
         _status.IsDataFromAux = false;
 
-        _bufferStartIdx = 0;
-        _bufferNumUsed = 0;
+        _buffer.Clear();
 
         bool shouldNotifyAux = !_shouldSkipDeviceNotify && !IsReadyForAuxFrame;
         bool shouldNotifyKbd = !_shouldSkipDeviceNotify && !IsReadyForKbdFrame;
@@ -277,7 +273,7 @@ public partial class Intel8042Controller : DefaultIOPortHandler {
     private void EnforceBufferSpace(int numBytes = 1) {
         ArgumentOutOfRangeException.ThrowIfGreaterThan(numBytes, BufferSize);
 
-        if (BufferSize < _bufferNumUsed + numBytes) {
+        if (BufferSize < _buffer.Count + numBytes) {
             LogWarnBufferIsFull();
             FlushBuffer();
         }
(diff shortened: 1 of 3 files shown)


src/Spice86.Core/Emulator/Devices/Input/Keyboard/Intel8042Controller.ConfigByte.cs
```diff
@@ -0,0 +1,57 @@
...
+public partial class Intel8042Controller {
+    /// <summary>
+    /// Byte 0x00 of the controller memory - configuration byte
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public sealed class ConfigByte {
+        public ConfigByte(byte initial = 0) => FromByte(initial);
+
+        // bit0: keyboard IRQ enabled (IRQ1)
+        public bool KbdIrqEnabled { get; set; }
```
kevinferrare, line 21: None of those properties look used.

I would invert the logic, ToByte / FromByte could return a stored byte and those properties could get / set the bit like it is done for cpu flags.

If needed, cpu flag code that does that could be extracted to a util since this is a common pattern

maximilien-noal: I looked into this, they seem to be referenced ?

image

maximilien-noal: I added all the references into the documentation.

They all have both Read and Write references in the controller.

src/Spice86.Core/Emulator/Devices/Input/Keyboard/Intel8042Controller.cs
```diff
@@ -0,0 +1,898 @@
...
+    private const byte FirmwareRevision = 0x00;
+    private const string FirmwareCopyright = nameof(Spice86);
+
+    private const int BufferSize = 64; // in bytes
+    /// <summary>
+    /// delay appropriate for 20-30 kHz serial clock and 11 bits/byte
+    /// </summary>
+    private const double PortDelayMs = 0.300; // 0.3ms
+
+    private readonly BufferEntry[] _buffer = new BufferEntry[BufferSize];
```
kevinferrare: Any reason for not using Queue<BufferEntry> ? This would make the intent clearer and code that manages the buffer simpler

maximilien-noal: None. It should be a queue or rather a circular buffer

src/Spice86.Core/Emulator/Devices/Input/Keyboard/PS2Keyboard.cs
```diff
@@ -0,0 +1,569 @@
...
+    private readonly DualPic _dualPic;
+    private readonly IGuiKeyboardEvents? _gui;
+    private readonly EmulatedTimeEventHandler _typematicTickHandler;
+    private readonly EmulatedTimeEventHandler _ledsAllOnExpireHandler;
+
+    /// <summary>
+    /// Internal keyboard scancode buffer
+    /// </summary>
+    private const int BufferSizeInScancodes = 8;
+    private readonly List<byte>[] _buffer = new List<byte>[BufferSizeInScancodes];
```
kevinferrare: Could this be a csharp queue? Maybe a better name would be _keyboardFrameBuffer?

maximilien-noal: Yes that would be better.

Although looking at the usage, isn't it a circular buffer ?

kevinferrare (approved):

PR #2014, opened 2026-03-22
feat: Live operand value display in disassembly view

maximilien-noal:
### Description of Changes

Shows register and memory operand values inline in the disassembly view when paused. Only visible lines are evaluated.

- Added `CompileValue` to `BreakpointConditionCompiler`to support value-returning expressions.
- New `ExpressionEvaluationService` maps Iced.Intel instruction operands to expression strings the AST parser understands, compiles them via `CompileValue`, and formats results as syntax-colored tokens matching the disassembly palette. Immediate and branch operands are skipped since they are already visible in the disassembly text.
- View tracks scroll position via `ScrollViewer.ScrollChanged` and reports the visible index range to the ViewModel. Evaluation runs on pause and on scroll while paused. Values are cleared on resume.
- `DebuggerLineViewModel` gets an `EvaluatedOperands` property, displayed in a new column in the disassembly grid.
- Renamed `FormattedTextSegmentsConverter` to `FormattedTextOffsetsConverter`.
- 7 new `CompileValue` unit tests covering register widths, segment registers, and memory reads. 4 new Avalonia headless UI tests for operand formatting.

### Rationale behind Changes

The disassembly view previously showed only instruction mnemonics and operand names. With this change, the actual runtime values of registers and memory locations referenced by each instruction are evaluated and displayed. This makes it possible to see what data an instruction is operating on without manually looking up each register or memory address.

### Suggested Testing Steps

Tested with Dune. Run `dotnet test tests/Spice86.Tests` for the new unit and UI tests. To test manually: pause the emulator, check that operand values appear next to instructions, scroll to see new lines get evaluated, and resume to confirm values are cleared.

image


maximilien-noal:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/OpenRakis/Spice86/pull/2014#pullrequestreview-3988458438)

src/Spice86.Core/Emulator/CPU/CfgCpu/InstructionExecutor/Expressions/AstExpressionBuilder.cs
```diff
@@ -209,6 +209,15 @@ public Expression VisitSegmentRegisterNode(SegmentRegisterNode node) {
     public Expression VisitSegmentedPointer(SegmentedPointerNode node) {
         Expression segmentExpression = node.Segment.Accept(this);
         Expression offsetExpression = node.Offset.Accept(this);
+        // The dual-parameter indexer expects (ushort segment, ushort offset).
```
kevinferrare: This is fixed (another way) in branch copilot/refactor-cfgcpu-execution-ast (this is actually a bug, code that produces segmented pointer should never produce anything else than  16 bits values)

src/Spice86.Core/Emulator/VM/Breakpoint/ExpressionValueCompiler.cs
```diff
@@ -0,0 +1,42 @@
...
+using Spice86.Core.Emulator.CPU.CfgCpu.InstructionExecutor.Expressions;
+using Spice86.Core.Emulator.Memory;
+using System.Linq.Expressions;
+
+/// <summary>
+/// Compiles expressions into value-returning functions using Expression trees.
+/// This is the value-evaluation counterpart to <see cref="BreakpointConditionCompiler"/>:
+/// same parser and builder infrastructure, but returns numeric values instead of booleans.
+/// </summary>
+public class ExpressionValueCompiler {
```
kevinferrare: This is also something that I did differently in the AST refactoring branch (because in the end, expressions need more than just state and memory):
```
        IVisitableAstNode executionAst = node.GenerateExecutionAst(astBuilder);
        AstExpressionBuilder expressionBuilder = new();
        CfgNodeExecutionAction<InstructionExecutionHelper> compiled =
            expressionBuilder.ToActionWithHelper(executionAst.Accept(expressionBuilder)).Compile();
```

Could be combined with BreakpointConditionCompiler which does sensibly the same thing

src/Spice86/ViewModels/Services/ExpressionEvaluationService.cs
```diff
@@ -0,0 +1,246 @@
...
+using Spice86.Core.Emulator.VM.Breakpoint;
+using Spice86.ViewModels.TextPresentation;
+
+/// <summary>
+/// Service for evaluating instruction operands against the current CPU state and memory.
+/// Extracts register and memory operands from Iced.Intel instructions, compiles them
+/// using the same expression infrastructure as breakpoint conditions, and formats the results
+/// as syntax-colored segments reusing the disassembly's <see cref="FormatterTextKind"/> palette.
+/// </summary>
+public class ExpressionEvaluationService {
```
kevinferrare, line 16: This would go away if ICED was replaced with our AST :)

maximilien-noal: Will do !

src/Spice86.Core/Emulator/VM/Breakpoint/ExpressionValueCompiler.cs
maximilien-noal: This class ExpressionValueCompiler  has been removed now.

kevinferrare (approved):

PR #2075, opened 2026-04-12
feat: Configurable emulator start time

kevinferrare:
### Description of Changes
Allow changing the start date of the emulator. This helps getting reproducible CPU logs.
This also helps if an application didnt expect to be started so far in the future :)


maximilien-noal (approved):
To ensure UTC, datimeoffset replaced datetime.

https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset?view=net-10.0

Datetime can have kind Local or Unknown.

If datetime has kind RTC, then it is 'nice'.

Otherwise it isn't very useful.

Datetimeoffset lifts that ambiguity by always being UTC.

But it's only my opinion. We already use datetime in the code base anyway, IIRC.

PR-2075-review-comments.md
maximilien-noal, line 1: Markdown file to be removed I think ?

maximilien-noal (changes requested):
LGTM, those tests are very useful. :)

Just a markdown file to remove I think.

maximilien-noal (approved):

PR #2161, opened 2026-05-07
Improvements to DosDriveManager

assumenothing:
### Description of Changes
Almost complete rewrite of `DosDriveManager`. Now using `DosDriveBase` as the base value type for the dictionary and unified all DOS drive types into a single internal collection type via an array. Created helper classes/structures inside to process the `Keys` and `Values` properties on the dictionary as well as fast `GetEnumerator()` support (most of that support code was inspired by the .NET `Dictionary<TKey,TValue>` implementation).

Added several helper methods into `DosDriveManager` for converting between DOS drive letters and zero-based indexes. I thought about putting them into `DosPathResolver` instead, but that is an internal class and these methods more closely related to drive management anyways.

Removed `DosDriveManager.DriveLetters` dictionary as it's no longer needed (and is likely very slow compared to the new methods).

Added methods for mounting, unmounting (including `IDisposable` and `IAsyncDisposable`), and getting DOS drives. (Some of the explicit/"hidden" dictionary methods do not dispose of drives that are removed--see documentation comments for applicable methods. Though this can be changed if it's more desirable to use dispose on anything that removes from the dictionary--I just wanted a way for advanced users to be able to remove specific drives from the drive manager without automatically calling `Dispose()` or `DisposeAsync()`.)

Moved `CurrentDosDirectory` property from `VirtualDrive` to `DosDriveBase` and made it optional (so it's not necessary to set it--the default value of an empty string should be sufficient).

Fixed LINQ calls to `ElementAtOrDefault` and `ElementAt` into `DosDriveManager`; those LINQ calls incorrectly assume that the enumerated element index is the same as the drive letter/index (which it's not--it just happens that drives A:, B:, and C: are all allocated and map to indexes 0, 1, and 2 respectively... other drives like Z: or any other user-mounted drives may not follow that pattern with the previous implementation).

Patched one of the tests so it wouldn't fail (see notes in "Suggested Testing Steps").

***NOTE:*** One thing that I have not done that I was thinking about doing: creating a new property called `DriveIndex` into `DosDriveBase` (with defensive bounds checks on the setter) and then redirecting the `DriveLetter` property to use that instead (via the new helper APIs I have created). Doing this would improve the reliability contract for the property, would remove the need to write defensive checks when accessing the `DriveLetter` property, and could also slightly improve performance by being able to access the drive index directly.

### Rationale behind Changes
This provides a better contract for future changes. It should be cleaner than the "multiple dictionary" approach that I've seen in other pull requests. There is only one drive letter to one drive, it doesn't make sense to me to keep the different drive types separated.

This should be faster than using `Dictionary<TKey,TValue>` fields as most operations are going to be retrieving values with specific drive letters and indexes, which have a one-to-one mapping. Array should be faster at doing this at a slight (probably imperceptible) performance cost when enumerating objects (which is not used often).

### Suggested Testing Steps
I have run all Spice86.Tests, but not Spice86.Tests.UI. There are currently 2 test failures and 1 skipped. I had patched another test, because it appears that the Z: drive is mapped by default in the unit tests (as a memory drive) and the specific test was using that drive letter to check for an "invalid drive" (and it no longer is with these changes).

Test failures are here: Spice86.Tests.Dos.DosConsoleIoctlIntegrationTests.Ioctl_OutOfScopeSubfunctions_ReturnConsistentFunctionInvalidError (for 15 and 17).
I'm not entirely sure if these failures are related or not, I'm not an expert at DOS system calls.

There are some path parsing bugs at present (especially in `DosPathResolver`), but I think those changes are better suited for a different pull request.

### DIFF
--- a/src/Spice86.Core/Emulator/OperatingSystem/DosDriveManager.cs
+++ b/src/Spice86.Core/Emulator/OperatingSystem/DosDriveManager.cs
@@ -339,10 +339,11 @@ public class DosDriveManager : IDictionary<char, DosDriveBase>, IReadOnlyDiction
         Mount(item.Value);
     }
 
-    /// <inheritdoc/>
+    /// <summary>Removes all mounted drives from the collection.</summary>
     /// <exception cref="AggregateException">One or more exceptions were thrown while drives were being unmounted.</exception>
     /// <remarks>
-    /// This will always dispose of the drives before removing them from the collection.
+    /// This will always dispose of the drives before removing them from the collection. (Equivalent to calling
+    /// <see cref="Clear(bool)"/> with the dispose drives parameter set to <see langword="true"/>.)
     /// </remarks>
     public void Clear() => Clear(disposeDrives: true);
 
@@ -778,8 +779,18 @@ public class DosDriveManager : IDictionary<char, DosDriveBase>, IReadOnlyDiction
 
     #endregion
 
+    /// <summary>Removes all mounted drives from the collection.</summary>
+    /// <param name="disposeDrives">
+    /// If <see langword="true"/>, then all currently mounted drives will be disposed (if applicable); otherwise,
+    /// drives will only be removed from the collection.
+    /// </param>
     /// <exception cref="AggregateException">One or more exceptions were thrown while drives were being unmounted.</exception>
-    public void Clear(bool disposeDrives = true) {
+    /// <remarks>
+    /// Removing drives from this collection without disposing may result in unexpected behavior or memory leaks. It is
+    /// up to the caller to make sure that any drives that are currently mounted are disposed before clearing the
+    /// collection.
+    /// </remarks>
+    public void Clear(bool disposeDrives) {
         if (!disposeDrives) {
             Array.Clear(_driveMap);
             _mappedDriveCount--;
--- a/src/Spice86.Core/Emulator/OperatingSystem/DosDriveManager.cs
+++ b/src/Spice86.Core/Emulator/OperatingSystem/DosDriveManager.cs
@@ -95,6 +95,17 @@ public class DosDriveManager : IDictionary<char, DosDriveBase>, IReadOnlyDiction
     }
 
     /// <summary>
+    /// Attempts to get the zero-based drive index associated with the given DOS drive letter.
+    /// </summary>
+    /// <param name="value">The DOS drive letter. Valid drive letters are uppercase and lowercase ASCII letters.</param>
+    /// <param name="driveIndex">The zero-based index associated with the drive letter or -1 on failure.</param>
+    /// <returns><see langword="true"/> if the drive letter and associated drive index is valid; otherwise, <see langword="false"/>.</returns>
+    public static bool TryGetDriveLetterIndex(char value, out int driveIndex) {
+        driveIndex = GetDriveLetterIndex(value);
+        return driveIndex != -1;
+    }
+
+    /// <summary>
     /// Gets the DOS drive letter from a zero-based drive index.
     /// </summary>
     /// <param name="index">Must be a zero-based drive index between 0 (inclusive) and <see cref="MaxDriveCount"/> (exclusive).</param>
@@ -182,7 +193,7 @@ public class DosDriveManager : IDictionary<char, DosDriveBase>, IReadOnlyDiction
     /// </summary>
     public byte CurrentDriveIndex => (byte)GetDriveLetterIndexOrThrow(CurrentDrive.DriveLetter);
 
-    internal bool HasDriveAtIndex(ushort zeroBasedIndex) => zeroBasedIndex is >= 0 and < MaxDriveCount &&
+    internal bool HasDriveAtIndex(int zeroBasedIndex) => zeroBasedIndex is >= 0 and < MaxDriveCount &&
         _driveMap[zeroBasedIndex] is not null;
 
     /// <summary>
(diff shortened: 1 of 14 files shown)


src/Spice86.Core/Emulator/OperatingSystem/DosDriveManager.cs
```diff
@@ -134,57 +229,785 @@ private byte MediaDescriptor(byte driveIndex) {
...
+        int itemCount = _mappedDriveCount;
+        if (array.Length - arrayIndex < itemCount) {
+            throw new ArgumentException("Destination array is not long enough to copy all the items in the collection. Check array index and length.");
+        }
+
+        DosDriveBase?[] entries = _driveMap;
+        for (int i = 0; i < MaxDriveCount; i++) {
+            DosDriveBase? entry = entries[i];
+            if (entry is not null) {
+                Debug.Assert(arrayIndex < itemCount);
```
assumenothing: This assertion is problematic if the caller does not use an array index of zero.

src/Spice86.Core/Emulator/OperatingSystem/DosDriveManager.cs
```diff
@@ -134,57 +229,785 @@ private byte MediaDescriptor(byte driveIndex) {
...
+            int itemCount = Count;
+            if (array.Length - arrayIndex < itemCount) {
+                throw new ArgumentException("Destination array is not long enough to copy all the items in the collection. Check array index and length.");
+            }
+
+            DosDriveBase?[] entries = _dictionary._driveMap;
+            for (int i = 0; i < itemCount; i++) {
+                DosDriveBase? entry = entries[i];
+                if (entry is not null) {
+                    Debug.Assert(arrayIndex < itemCount);
```
assumenothing: This assertion is problematic if the caller does not use an array index of zero.

src/Spice86.Core/Emulator/OperatingSystem/DosDriveManager.cs
```diff
@@ -134,57 +229,785 @@ private byte MediaDescriptor(byte driveIndex) {
...
```
assumenothing: This assertion is problematic if the caller does not use an array index of zero.

assumenothing:
I think I fixed all the bugs/regressions in this pull request, as far as I know. The tests now appear to be passing.

maximilien-noal (approved):

maximilien-noal:
@assumenothing I approved the PR. Thank you so much for contributing. :)

I really like the new API.

Just two things:

- Can Debug.Assert be replaced by exceptions or be removed ?
Debug.Assert can trigger in Release builds, and if it triggers the normal path for an unhandled exception is not executed. Unhandled exceptions are reporte, ,files are closed, dumps are produced, etc...

-  Can `#region ` directives be removed ?

assumenothing:
@maximilien-noal 
> * Can Debug.Assert be replaced by exceptions or be removed ?
>   Debug.Assert can trigger in Release builds, and if it triggers the normal path for an unhandled exception is not executed. Unhandled exceptions are reporte, ,files are closed, dumps are produced, etc...

`Debug.Assert` uses the `ConditionalAttribute` with a test for the "DEBUG" compiler symbol. So the C# compiler will automatically remove calls to methods marked with that attribute when the symbol "DEBUG" is not defined (which is not defined in release builds for Spice86). Thus they will not trigger in release builds, only `Trace.Assert` calls will trigger in release builds (because release builds define the "TRACE" compiler symbol).

I believe all the code paths where I added debug assertions are only for extra sanity when debugging and are used to potentially identify problems early before other exceptions are eventually thrown (like `NullReferenceException`, `IndexOutOfBoundsException`, etc). There shouldn't be a performance degradation in release builds when using them and they won't throw an exception or trigger in release builds.

They are often used as a type of contract that the compiler can use to determine when a variable may be not null (otherwise ugly hacks like `!` operators have to be used). They can also be used for performance critical code where there are "fast" paths which ideally shouldn't throw exceptions (throwing and catching exceptions can sometimes slow down the runtime with the .NET JIT & AOT engines--if I remember correctly, methods and properties which use `throw` and `try..catch..finally` blocks currently cannot be inlined at runtime). The official .NET source code uses debug assertions all over the place for some of the same reasons I have given above (it is one of the main reasons why it exists).

But I can remove/refactor them if you believe that they should not be used in this project.

> * Can `#region ` directives be removed ?

Yes, I can remove those directives. I was only using them in the source to make it a little bit easier to identify what "category" the code belongs to.

I also found a bug in `DriveLetterCollection.GetEnumerator()` and `DriveCollection.GetEnumerator()` where the last element (Z: drive) is zero/null. Though for some currently unknown reason, `DosDriveManager.GetEnumerator()` appears to be working as intended. I'm working on figuring out why and try to fix (the code between all three enumerator types are nearly identical).

maximilien-noal:
@assumenothing 

You're right, indeed .NET uses it all over the place.

I'll wait on your feedback before merging, then. Hope you'll find the reason for this bug.

assumenothing:
@maximilien-noal I figured out why it wasn't working. It was an issue I missed in some of the `CopyTo` methods, not the `GetEnumerator` methods. I had to load up the compiled assemblies in ILspy to figure out why it was working for some tests and not for others or the debugger visualizer.

I added some basic unit tests for `DosDriveManager`. That should help with identifying any regressions in the future.

_On another note: once these changes are merged, I have another set of changes in another branch on my Git repository that I have been working on relating to fixing the DOS path resolver and a few bugged unit tests. But that should be put into another pull request, I think. [fix-dos-path-resolver branch](https://github.com/assumenothing/Spice86/tree/fix-dos-path-resolver)_

maximilien-noal (approved):

maximilien-noal:
@assumenothing Great ! Thank you. 😊

PR Merged.

PR #2168, opened 2026-05-12
fix: DISASM view is too slow

kevinferrare:
### Description of Changes
Disassembly view was taking 20sec to show something on my machine that is not a dinosaur.


maximilien-noal (approved):

maximilien-noal:
Looks good to me !

PR #2181, opened 2026-05-18
refactor: Optimized & improved DosExecResult and DosFileOperationResult.

assumenothing:
### Description of Changes
- `DosExecResult` now implements `IEquatable<T>` interface and overrides `Equals(object?)` and `GetHashCode()`.
- `DosExecResult` added static cache for storing a subset of common error results.
- `DosFileOperationResult` now implements `IEquatable<T>` interface and overrides `Equals(object?)` and `GetHashCode()`.
- `DosFileOperationResult` added static cache for storing a subset of common error results, `NoValue()` result, and repeated `Value16()` results (with the same value).
- `DosFileOperationResult` class is now sealed (was not possible to inherit before with private constructor anyways).
- Added parameter names to a few calls to make it more obvious what parameter values are being passed into the constructors.

### Rationale behind Changes
Minor performance improvements for fixed/common/repeated operation results (reduces heap allocations).

It would also be possible to easily cache the fixed result from `DosExecResult.SuccessLoadOverlay()`, but this is most likely not called often enough to need that optimization.

**NOTE:** It would be better/more efficient to change `DosExecResult` and `DosFileOperationResult` from `class` (reference type) to `readonly struct` (value type), as there would then be no need to apply these caching optimizations. (This would further reduce heap allocations, may slightly improve memory accesses as the results would generally be stored in the thread stack rather than the GC heap, would potentially avoid an extra pointer dereference, and would have the minor benefit of being able to safely implement the `==` and `!=` operators). However, these are public APIs and changing the fundamental object types would be a breaking change.

### Suggested Testing Steps
Tested against project unit tests.

### DIFF
--- a/src/Spice86.Core/Emulator/OperatingSystem/Structures/DosExecResult.cs
+++ b/src/Spice86.Core/Emulator/OperatingSystem/Structures/DosExecResult.cs
@@ -40,8 +40,11 @@ public sealed class DosExecResult : IEquatable<DosExecResult?> {
     public static DosExecResult Fail(DosErrorCode code) {
         // Use cache for common error codes.
         if ((int)code is >= 0 and < ErrorCacheLength) {
-            ref DosExecResult? errorCacheEntry = ref s_errorCache[(uint)code];
-            errorCacheEntry ??= NewErrorResult(code);
+            DosExecResult? errorCacheEntry = s_errorCache[(int)code];
+            if (errorCacheEntry is null) {
+                errorCacheEntry = NewErrorResult(code);
+                s_errorCache[(int)code] = errorCacheEntry;
+            }
             return errorCacheEntry;
         }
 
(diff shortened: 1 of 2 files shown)


assumenothing:
I did think about the possible thread safety issues (and fixed one of the *theoretically possible* threading issues I found in my first commit). But it shouldn't really matter whether the caches thread-safe or not: they are immutable objects and the static fields references are copied into local variables before accessing them. I could add the `volatile` keyword to the fields or use the `Volatile` class methods, but again it shouldn't really make much difference (some duplicate instances are better than a lot of duplicates like it was prior to these changes--these are small objects anyways).

maximilien-noal:
@assumenothing This is a natural evolution of those classes, and yeah they should have been sealed long ago.

I'm always on the side of more performance, especially since long GC pauses are problematic with sound rendering.

Approved! :)

maximilien-noal (approved):

PR #2184, opened 2026-05-20
enhancement: Provide optimized Span-based memory access APIs.

assumenothing:
### Description of Changes
I started with virtual method additions to `IReaderWriter<T>` to be able to request `Span<T>` and `ReadOnlySpan<T>` instances (with memory access modifiers). This propagated into changes to `ArrayReaderWriter`, `ByteArrayReaderWriter`, `UintArrayReaderWriter`, and `ByteReaderWriterWithBaseAddress`. Some of these classes were changed from abstract to non-abstract and others were sealed to try to maximize performance by allowing the runtime to devirtualize some of these classes (though it's difficult to determine if that made any performance improvements).

Then I updated the various `...Indexer` classes to utilize these new span-based APIs when possible (except for `Int8Indexer` and `UInt8Indexer` as there is no benefit to those). Some of the optimizations in these might be a bit aggressive, but the intent is to reduce as much overhead from runtime-generated bounds checks where it is probably difficult for the runtime to optimize away. Though the fallback non-span implementations are mostly left untouched.

And finally I updated the `Indexable` abstract class (and the related `IIndexable` interface) and added a few span-based helpers as well as utilizing the new enhancements to `IReaderWriter<T>`. I also made most of the methods virtual to allow for more optimized span-based implementations to be used where appropriate. Text encoding uses `Encoding.Latin1` where appropriate/useful (this encoding is equivalent to casting all `byte` values 0-255 directly to `char` but utilizes a replacement character when going in reverse for values outside this range). This propagated into potential performance enhancing changes to `ByteArrayBasedIndexable` and `AbstractMemoryBasedDataStructure`.

- `Indexable`: Fixed an overlapping memory bug in `MemCopy` implementation. This method is now roughly comparable to the C function `memmove` instead of `memcpy` to avoid possible memory corruption issues.

- [ ] **TODO:** I still need to write some new unit tests for these new & modified APIs just in case any difficult to identify bugs may be lurking.

### Rationale behind Changes
Should open the door for further performance enhancements to the memory subsystem.

Eventually the idea is to be able to retrieve spans and perform faster operations that aren't directly affected by debugger memory breakpoints for more than single byte memory accesses through `IMemory` and `IMemoryDevice`, but I feel like that requires updating some of these low-level access classes & interfaces first.

Right now I don't believe there is a significant performance increase with these changes. (In theory there might actually be a slight slowdown right now from the span checks on `short`, `ushort`, `int`, and `uint` accesses, but I have not observed this from running the unit tests.) But I think with some other changes, either on this draft pull request or another one, this may improve.

### Suggested Testing Steps
I believe this currently passes all unit tests. I have not tested with any real world DOS programs.


assumenothing:
Other than needing to write some unit tests to check for some edge cases and standard usage patterns, I think the code updates are nearly ready (unless I find more bugs).

I found a few methods in `Indexable` which didn't have any documentation or explicit code for handling out of range length values:
- `string GetZeroTerminatedString(uint address, int maxLength)`
  - What happens when maximum length is negative? ignore?
  - Documentation should probably indicate that it will truncate the string to the maximum length.
- `void SetZeroTerminatedString(uint address, string value, int maxLength)`
  - What happens when maximum length is negative? ignore?
  - Should this truncate the string with a NUL terminator if it's too long? Currently throws `UnrecoverableException`.
- `string GetSpacePaddedString(uint address, int length)`
  - What happens when length is negative? ignore?
- `void SetSpacePaddedString(uint address, string value, int length)`
  - What happens when length is negative? ignore?
- `void LoadData(uint address, byte[] data, int length)`
  - What happens when length is negative or greater than `data.Length`? ignore?
- `void LoadData(uint address, ushort[] data, int length)`
  - What happens when length is negative or greater than `data.Length`? ignore?

kevinferrare:
@assumenothing Maybe I didnt get what you want to do, but I dont think you can use spans to represent memory, it's a device mapper, each address potentially belongs to a device, it's not a linear array.
Indexers on actual arrays will maybe be faster with spans but that's an edge case.

assumenothing:
> @assumenothing Maybe I didnt get what you want to do, but I dont think you can use spans to represent memory, it's a device mapper, each address potentially belongs to a device, it's not a linear array. Indexers on actual arrays will maybe be faster with spans but that's an edge case.

@kevinferrare I did see that's how the Memory class is currently implemented, but it is possible to optimize it to retrieve spans of memory, if a given `IMemoryDevice` is capable of utilizing span-based APIs (similar to the methods I added to `IReaderWriter<T>`).

But there are still a few non-span related `IMmu` optimizations that I think have yielded a small performance gain (by reducing multiple virtual calls to `TranslateAddress` into a single `TryTranslateAddressRange` virtual call--that try method currently only fails when the memory has to wrap around a segment offset boundary).

One potential optimization I was thinking about (possibly in the future) was to turn that big `IMemoryDevice` array field inside `Memory` into a series of fixed-sized "super blocks" that can be configured to quickly indicate whether the super block is mapped to a single device (with a single nullable `IMemoryDevice` field) or mapped to several (possibly using a linked list of mapped memory ranges and/or just another smaller array in each super block). And obviously some checks would need to be added to make sure breakpoints are not enabled on these addresses when trying to get spans of read/write sensitive memory.

kevinferrare:
> > @assumenothing Maybe I didnt get what you want to do, but I dont think you can use spans to represent memory, it's a device mapper, each address potentially belongs to a device, it's not a linear array. Indexers on actual arrays will maybe be faster with spans but that's an edge case.
> 
> @kevinferrare I did see that's how the Memory class is currently implemented, but it is possible to optimize it to retrieve spans of memory, if a given `IMemoryDevice` is capable of utilizing span-based APIs (similar to the methods I added to `IReaderWriter<T>`).
> 
> But there are still a few non-span related `IMmu` optimizations that I think have yielded a small performance gain (by reducing multiple virtual calls to `TranslateAddress` into a single `TryTranslateAddressRange` virtual call--that try method currently only fails when the memory has to wrap around a segment offset boundary).
> 
> One potential optimization I was thinking about (possibly in the future) was to turn that big `IMemoryDevice` array field inside `Memory` into a series of fixed-sized "super blocks" that can be configured to quickly indicate whether the super block is mapped to a single device (with a single nullable `IMemoryDevice` field) or mapped to several (possibly using a linked list of mapped memory ranges and/or just another smaller array in each super block). And obviously some checks would need to be added to make sure breakpoints are not enabled on these addresses when trying to get spans of read/write sensitive memory.

@assumenothing Thanks for the detailed explanation!

The main access patterns are the instructions writing byte / word / dword (8 / 16 / 32) and the rep prefixed string operations. Other access patterns are very infrequent and probably not worth optimizing.

Someone in the past tried to replace some of the REP accesses with spans and a check on:
- the segment bounds (similar to your TryTranslateAddressRange)
- the device bounds (with blocks spanning between devices, if block for start and end address changed, it would not be possible to use span).

if all OK (start / end within segment and device block bound) -> access directly the device memory.
However, this was a lot of complexity added for a small boost (he had to implement it basically for all string instructions a bit differently with code for span and a code for simple access). On top of that, we moved to an AST based execution engine to enable code generation from execution trace, so if AST for every rep instruction doubles in size it will make generated code bigger as well.

As of now, do you observe some performance gain with your PR?

One thing that would bring benefits without too much pain would be the breakpoints as you identified.
I think it would be even simpler (and faster?) to have one struct per memory address with redirection to the device and flags saying what kind of breakpoint is there (read / write), and if there is and it matches the operation, we activate the breakpoint check as we do now, if not, just access.

Also I am curious to hear what brought you to Spice86, it's a fairly niche project😄

assumenothing:
@kevinferrare Thanks for responding and providing some insight into previous optimization attempts.

It's difficult to judge how much time is being saved by my optimizations, because the unit tests will execute on my system with a variance of approximately ±30 seconds to ±1 minute when using Visual Studio's timing information (even when using a ramdisk for %TEMP% and clearing it before each run). I felt like I was making a little bit of progress on the tests, but I have been experimenting today with trying to make a basic span access integration in the `Ram` class. but it is currently very naive and slow, with no real noticeable performance gains at this point (and Visual Studio doesn't want to run with profiling turned on with those unit tests for some unknown reason).

It might still be a small performance win if the breakpoint checks and device address range checks could be optimized--though the breakpoints are currently implemented in a way that makes it challenging to synchronize into `Memory` via a bitmask or some other cache (it would require some internal properties to tie breakpoints into `Memory` for automatic updates when the `IsEnabled` flag changes and there are also some challenges with unconditional breakpoints too which affect all memory addresses from all appearances).

As you said though, it might be adding too much complexity for very small possible performance improvements. Though maybe it would help with the more complex cases like DOS emulation and reverse engineering where there is more structured memory access involved with larger reads & writes.

I found this project several years ago when I was trying to reverse engineer an old proprietary and niche scientific DOS program at work (it was apparently built using the ancient Microsoft FORTRAN compiler of all things). So I wanted to find a decent C#-written/.NET compatible x86 DOS emulator and some tools to go with it (since I'm very familiar with .NET & C#--I've been writing code for it ever since the original .NET Framework 1.1/2.0 days and Visual Studio 2005). Unfortunately, the program I was trying to reverse engineer uses the x87 coprocessor instructions (and didn't compile in support for any of the floating-point emulation libraries that were around at the time). So instead I had to reverse engineer it the hard way with IDA and Ghidra and rewrote it in C++. So I still follow this project here and there just for fun as it looks interesting.

kevinferrare:
> @kevinferrare Thanks for responding and providing some insight into previous optimization attempts.
> 
> It's difficult to judge how much time is being saved by my optimizations, because the unit tests will execute on my system with a variance of approximately ±30 seconds to ±1 minute when using Visual Studio's timing information (even when using a ramdisk for %TEMP% and clearing it before each run). I felt like I was making a little bit of progress on the tests, but I have been experimenting today with trying to make a basic span access integration in the `Ram` class. but it is currently very naive and slow, with no real noticeable performance gains at this point (and Visual Studio doesn't want to run with profiling turned on with those unit tests for some unknown reason).
> 
> It might still be a small performance win if the breakpoint checks and device address range checks could be optimized--though the breakpoints are currently implemented in a way that makes it challenging to synchronize into `Memory` via a bitmask or some other cache (it would require some internal properties to tie breakpoints into `Memory` for automatic updates when the `IsEnabled` flag changes and there are also some challenges with unconditional breakpoints too which affect all memory addresses from all appearances).
> 
> As you said though, it might be adding too much complexity for very small possible performance improvements. Though maybe it would help with the more complex cases like DOS emulation and reverse engineering where there is more structured memory access involved with larger reads & writes.
> 
> I found this project several years ago when I was trying to reverse engineer an old proprietary and niche scientific DOS program at work (it was apparently built using the ancient Microsoft FORTRAN compiler of all things). So I wanted to find a decent C#-written/.NET compatible x86 DOS emulator and some tools to go with it (since I'm very familiar with .NET & C#--I've been writing code for it ever since the original .NET Framework 1.1/2.0 days and Visual Studio 2005). Unfortunately, the program I was trying to reverse engineer uses the x87 coprocessor instructions (and didn't compile in support for any of the floating-point emulation libraries that were around at the time). So instead I had to reverse engineer it the hard way with IDA and Ghidra and rewrote it in C++. So I still follow this project here and there just for fun as it looks interesting.

@assumenothing For the Span that's my feeling, I don't remember the figures but even back then when code was simpler, gain was too small to bother with the the added layer of complexity, and it introduced bugs. I forgot to mention that rep writes must trigger breakpoints for the addresses being accessed (we need the breakpoints to detect self modifying code).

For the breakpoint optimization, unconditional ones are an issue indeed.
We could unplug them (I think only UI sets them and I never used them) or just check that none of them is enabled per access (they are a rare case).
Then the check would be for the specific memory location being accessed without going through a dictionary. Would be a win to get rid of that.
For IsEnabled, even if we check disabled breakpoints it would be better than what we do now (check them for every memory access).
There are many breakpoints enabled behind the scene so this would have an impact. This is because this is how the CPU detects self modifying code and invalidate caches to rewire the CFG graph we thread as the execution goes on (graph is visible via the UI if you saw).

I'm happy you find this project interesting. I had to fiddle with ghidra in the past (wrote a plugin) and it was a disaster with 16bit segmented x86 code. Must have been a pain to reverse your program. I head IDA is better but hex-rays doesnt work for 16bits.

On this topic, there is a code generator v2 coming (one that should work even for crazy cases and not depend on ghidra).

assumenothing:
@kevinferrare Yeah, from all of my testing so far, there seems to be virtually no difference in performance of the unit tests between when using the basic span optimizations in this pull request and without them. It seems about the same even when I bypass some of the more expensive checks in `Memory.TryGetSpan` (like breakpoints and memory devices--obviously breaking a bunch of tests in the process).

It's possible that real applications might perform slightly better, but I don't have any test applications set up for that kind of testing right now.

Optimizing the breakpoints inside `Memory` looks even more challenging, because apparently there can be custom trigger conditions set up on each per-address memory breakpoint.

It seems like if a complete breakpoint redesign were to be done, a better approach would be to just use a bitmap on every memory address for read breakpoints and another bitmap for write breakpoints and then set up breakpoints by changing the read/write flags in the memory. Then it would be up to the user to add events to handle the breakpoints more generically by testing the triggered memory address (I'm assuming that there aren't a _huge_ number of breakpoints being used at any given time). And for breakpoints that handle a larger range of addresses then you could adopt a Windows-like approach of setting read/write triggers for entire "pages" of memory (basically just a bitmap of breakpoints for every e.g. 64, 512, 4096, whatever number of bytes that defines a "page"). If I remember correctly, I believe this is how memory breakpoints are typically implemented on Windows (there are a fixed number of CPU hardware address breakpoints available and then beyond that you can also set up per-page one-shot read/write/execute triggers that the debugger can handle).

Below is the test code that I had added into `Memory` for expanded span testing (I also added the `TryGetSpan` methods to `IMemoryDevice` and implemented them in the `Ram` class). Note that I had to make a few changes to breakpoints to make this work (temporarily added `HasActiveBreakPoint` which just checks if `IsEnabled` is true on any applicable breakpoints for the given memory address or unconditional breakpoints).
```csharp
    public bool TryGetSpan(uint startAddress, int length, out Span<byte> span, MemoryAccess access = MemoryAccess.None) {
        ArgumentOutOfRangeException.ThrowIfNegative(length);

        // Transform start and end addresses through A20 gate.
        // Note that the end address is inclusive (to avoid issues with wrapped addresses on the last byte).
        startAddress = A20Gate.TransformAddress(startAddress);
        uint endAddress = A20Gate.TransformAddress(startAddress + (uint)length - 1);

        return TryGetSpan(startAddress, endAddress, out span, access);
    }

    public bool TryGetSpan(uint startAddress, int length, out ReadOnlySpan<byte> span, MemoryAccess access = MemoryAccess.None) {
        bool result = TryGetSpan(startAddress, length, out Span<byte> mutableSpan, access);
        span = mutableSpan;
        return result;
    }

    private bool TryGetSpan(uint startAddress, uint endAddress, out Span<byte> span, MemoryAccess access) {
        int length = (int)(endAddress - startAddress + 1);

        // Return an empty span if length is zero.
        if (length == 0) {
            span = [];
            return true;
        }

        IMemoryDevice?[] memoryDevices = _memoryDevices;

        // Cannot get a span if address overflowed/wrapped (span length is limited to a 32-bit signed integer anyways).
        // Also fail if end address exceeds total allocated device memory.
        if (length < 0 || endAddress >= (uint)memoryDevices.Length) {
            span = [];
            return false;
        }

        // Make sure entire memory range is associated with a single device.
        IMemoryDevice? device = memoryDevices[startAddress];
        if (device is null) {
            span = [];
            return false;
        }

        // TODO: Check breakpoint on first address before checking all devices?

        // TODO: This is slow, it would be better to use a linked list of device memory address ranges.
        for (int i = 1; i < length; i++) {
            if (memoryDevices[startAddress + (uint)i] != device) {
                span = [];
                return false;
            }
        }

        // Try to get the span from the device before the expensive breakpoint check.
        if (!device.TryGetSpan(startAddress, length, out span, access)) {
            span = [];
            return false;
        }

        // Check for breakpoints if memory access is requesting read or write. Ignore breakpoints if no access flags
        // are requested.
        access &= MemoryAccess.ReadWrite; // ignore all other flags
        if (access != MemoryAccess.None) {
            AddressOperation breakPointTrigger = access switch {
                MemoryAccess.Read => AddressOperation.READ,
                MemoryAccess.Write => AddressOperation.WRITE,
                _ => AddressOperation.ACCESS,
            };

            // TODO: This is slow, it would be nice if there was a cache/lookup table of active break points.
            // TODO: This does not check "dynamic" breakpoints that have custom trigger conditions defined.
            for (int i = 0; i < length; i++) {
                if (_memoryBreakpoints.HasActiveBreakPoint(startAddress + (uint)i, breakPointTrigger)) {
                    span = [];
                    return false;
                }
            }
        }

        // A valid span from a device with no set breakpoints should be a success.
        return true;
    }
```

assumenothing:
@kevinferrare @maximilien-noal I think this will probably be the last commit on this pull request. After experimenting with enabling and using spans with `Memory`, I haven't been able to improve the performance of the unit tests--probably due to heavy use of breakpoints as well as requiring too many defensive checks and layering required to successfully retrieve the spans.

I did find a number of minor bugs and a few other small features that might yield some small performance boosts, but the main span-related APIs don't seem to be all that useful for the amount of complexity that I've added--especially with the main element access methods via `...Indexer` classes. Though maybe some of these APIs might be beneficial for certain large read/write calls like working with video memory, extended memory, or for advanced users who want to perform operations on large chunks of memory without doing it one byte at a time.

At this point I might recommend closing and not merging this pull request (unless you think these span-based access APIs might be useful in the future, just not useful for the `...Indexer` classes right now).

But if merged, I would probably recommend setting the internal `Spice86.Core.Emulator.Memory.Indexable.DisableSpanAccess` property to `true` to disable the attempted "micro optimizations" so everything uses byte-by-byte accesses by default. (There might potentially be some performance gains in the string/memory functions in `Indexable`, but it looks like those calls are very rare.)

_P.S. I'm not sure why the Github automated tests fail on Windows after the force push to the latest master commits. It doesn't fail on my computer when I run all the tests..._

kevinferrare:
> @kevinferrare @maximilien-noal I think this will probably be the last commit on this pull request. After experimenting with enabling and using spans with `Memory`, I haven't been able to improve the performance of the unit tests--probably due to heavy use of breakpoints as well as requiring too many defensive checks and layering required to successfully retrieve the spans.
> 
> I did find a number of minor bugs and a few other small features that might yield some small performance boosts, but the main span-related APIs don't seem to be all that useful for the amount of complexity that I've added--especially with the main element access methods via `...Indexer` classes. Though maybe some of these APIs might be beneficial for certain large read/write calls like working with video memory, extended memory, or for advanced users who want to perform operations on large chunks of memory without doing it one byte at a time.
> 
> At this point I might recommend closing and not merging this pull request (unless you think these span-based access APIs might be useful in the future, just not useful for the `...Indexer` classes right now).
> 
> But if merged, I would probably recommend setting the internal `Spice86.Core.Emulator.Memory.Indexable.DisableSpanAccess` property to `true` to disable the attempted "micro optimizations" so everything uses byte-by-byte accesses by default. (There might potentially be some performance gains in the string/memory functions in `Indexable`, but it looks like those calls are very rare.)
> 
> _P.S. I'm not sure why the Github automated tests fail on Windows after the force push to the latest master commits. It doesn't fail on my computer when I run all the tests..._

@assumenothing I think most of the time in the tests is spent on 386 single step steps, we execute ~1.2million tests where:
- we parse json for the test
- we execute the 2 or 3 instructions for that test
- we compare results with expected
- we reset the machine so that it is ready for next test
I am sure it could be done in a smarter way, but not all the time is spent in the execution engine for this particular test suite. If I remember well, json parsing alone was taking like 2 minutes. Maybe there is an easy win there?

assumenothing:
> I think most of the time in the tests is spent on 386 single step steps, we execute ~1.2million tests where:
> * we parse json for the test
> * we execute the 2 or 3 instructions for that test
> * we compare results with expected
> * we reset the machine so that it is ready for next test
>       I am sure it could be done in a smarter way, but not all the time is spent in the execution engine for this particular test suite. If I remember well, json parsing alone was taking like 2 minutes. Maybe there is an easy win there?

@kevinferrare I was typically skipping those 5 tests on my local machine because I knew that there was not going to be much hope of being able to optimize those via memory modifications (and they take around 5 to 10 minutes to execute on my machine).

As far as JSON parsing, I'm not entirely sure how it's currently being done, but I would think the best performance would probably be to use a low-level JSON reader like `System.Text.Json.Utf8JsonReader` to manually iterate through the JSON tokens.

So I ran the intro sequence for Space Quest 2 (from the Steam collection) and I got the following numbers from just running the intro sequence and idling for a minute or two:
| Configuration | `...Indexer` Spans | `Indexable` Spans | Debug Build | Release Build |
| --- | :---: | :---: | :---: | :---: |
| Master 3a3f0bd | n/a | n/a | 2 378 557 ips (100%) | 24 151 298 ips (100%) |
| PR #2184 | :x: Disabled | :x: Disabled | 2 349 589 ips (98.78%) | 23 756 911 ips (98.37%) |
| PR #2184 | :x: Disabled | :white_check_mark: Enabled | 2 265 776 ips (95.26%) | 23 180 926 ips (95.98%) |
| PR #2184 | :white_check_mark: Enabled | :white_check_mark: Enabled | 2 189 423 ips (92.05%) | 21 323 734 ips (88.29%) |

_"ips" is Average Instructions per Second according to statistics logging. Debug builds were run for approximately 250 000 000 instructions and Release builds were run for approximately 2 000 000 000 instructions. My development machine is currently using an AMD Ryzen 3900X with 64GB of 2933 MT/s memory (it's several years old now)._

And I was finally able to get Spice86 running with CPU profiling turned on. It appears that over 5% of the execution time is spent inside `UInt16Indexer.ReadSegmented` when spans are enabled and less than 1% when spans are disabled (this is most likely due to the expensive breakpoint checks being shifted from other APIs over to `ReadSegmented`). Around 5% of the execution time is dedicated to the `RealModeMmu386.CheckAccess` method (and even with my simplified expression there isn't a whole lot that can be reduced in that method). Interestingly one of the biggest hot paths Visual Studio was indicating for the demo program I was running is in `Spice86.Core.Emulator.CPU.Alu8.Sub(byte, byte, bool)` and `Spice86.Core.Emulator.CPU.Alu16.Sub(ushort, ushort, bool)` (around 2% for each method).

maximilien-noal:
@assumenothing I'll close this if you don't mind ?

assumenothing:
Yes, please close. I don't think there is really much I can do at this time to get the performance improvements I was looking for with these changes.

PR #2196, opened 2026-05-24
CD emulation, floppy emulation, CDDA support, floppy image support, CD images support, command line interpreter drive switching support, and DOS echo fix

maximilien-noal:
### Description of Changes

Brings full CD drive and floppy drive emulation and images support.

### Rationale behind Changes

Some games seem to require it. Some complain about the lack of CDs. Others seem to only run from floppy.

### Suggested Testing Steps

- Run `dotnet test`

All new integration tests should pass.


kevinferrare (approved):

src/Spice86.Core/Emulator/OperatingSystem/DosDriveManager.cs
```diff
@@ -995,4 +1004,425 @@ public bool TryGetDriveAtIndex<T>(int driveIndex, [NotNullWhen(true)] out T? val
...
+        }
+        _driveMap[idx] = newDrive;
+        _mappedDriveCount++;
+        _version++;
+    }
+
+    private bool TryResolveImageBackedDrive(byte driveNumber, out FloppyDiskDrive? floppy, out byte[]? imageData) {
+        floppy = null;
+        imageData = null;
+        char driveLetter = (char)('A' + driveNumber);
```
assumenothing: This and the range check below can be replaced with a call to `TryGetDriveLetterFromIndex(driveNumber)` inside a conditional.

Alternatively you could use `char driveLetter = GetDriveLetterFromIndexFast(driveNumber);`, remove the range check below and let the call to `TryGetFloppyDrive` automatically fail with the invalid drive letter if the drive number is out of range (this alternative method only works because "drive number" must be a value between 0 and 255 and adding "A" to it and casting to a UInt16 (char) will not lose any bits--if the parameter were a larger type like an "int", then `GetDriveLetterFromIndexFast` could break from lost bits during conversion).

Edit: though now that I'm thinking about it, `GetDriveLetterFromIndexFast` has a debug assertion tied to the input range, so `TryGetDriveLetterFromIndex` would be the preferred way to accomplish this using existing APIs.

src/Spice86.Core/Emulator/OperatingSystem/Structures/DosPathBuilder.cs
```diff
@@ -79,12 +72,22 @@ public DosPathBuilder(Span<char> scratchPathBuffer, Span<int> scratchStackBuffer
         _pathStack = new(scratchStackBuffer);
     }
 
-    /// <summary>Initializes a DOS path builder with array pool allocated buffers.</summary>
+    /// <summary>Creates an empty DOS path builder with no preallocated scratch buffers.</summary>
+    /// <remarks>
+    /// This is useful when calling <see cref="ParseSpecialFileName(ReadOnlySpan{char})"/> without building a DOS path.
+    /// </remarks>
+    public static DosPathBuilder Create() {
+        return new DosPathBuilder(Span<char>.Empty, Span<int>.Empty);
```
assumenothing: The same could be achieved by using `return new DosPathBuilder();`. `Span<char>.Empty` is basically the same as `default(Span<char>)` and `struct` types always have a default constructor that initializes all fields to `default`. (Only using `Unsafe.SkipInit<T>()` or other IL hacks can bypass that rule.)

maximilien-noal:
@copilot fix implementation by comparing with https://github com/FDOS/kernel so all dos tests pass again. All tests should pass with a minimum amount of changes. Do not add workarounds.

kevinferrare (approved):

PR #2217, opened 2026-06-06
fix: SetRTCDate uses SetCarryFlag

maximilien-noal:
### Description of Changes

Fixes tests with RTC clock INT1AH handler

### Rationale behind Changes

Oops, I broke them by correcting State.CarryFlag usage by SetCarryyFlag in the INT1AH interrupt handler.

While this was correct, tests did not like that.

### Suggested Testing Steps

All tests should pass.


maximilien-noal:
The macOS runner not passing is an Out of memory error I can't reproduce in DOS CHOICE integration tests.

kevinferrare (approved):

PR #2258, opened 2026-07-12
fix: Show 16 bytes per row in the memory view of the internal debugger

brunom:
Previously it showed 13 (or 0x13) columns which didn't make sense


maximilien-noal (approved):

maximilien-noal:
@brunom nice find, thanks! :)

PR #2259, opened 2026-07-12
fix: match .editorconfig to actual practice

brunom:
Now Visual Studio formatting leaves files unchanged. Tested on src/Spice86/ViewModels/DisassemblyViewModel.cs

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -130,12 +130,12 @@ csharp_style_allow_embedded_statements_on_same_line_experimental = true
 #### C# Formatting Rules ####
 
 # New line preferences
-csharp_new_line_before_open_brace = none
-csharp_new_line_before_catch = false
-csharp_new_line_before_else = false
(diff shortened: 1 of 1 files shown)


maximilien-noal:
@brunom would not this break the java style code blocks used in all files ?

brunom:
I found out .editorconfig didn't match DisassemblyViewModel.cs while fixing https://github.com/OpenRakis/Spice86/pull/2260
The Copilot reviewer here says other files use other styles. Maybe different .editorconfig files per folder or sub project would match the files on disk. Or we could reformat everything to one style. I just wanted to raise the issue.

maximilien-noal:
@brunom thanks for raising the issue, I wasn't aware of it. Will fix it sometime later with dotnet format.

Is it OK for you if this PR is closed ?

brunom:
Sure, thanks.

PR #2260, opened 2026-07-12
fix: Internal debugger crash when instruction would read 2 bytes starting at 0xFFFF

brunom:
When a function exit poped 0xffff into si, an instruction far away that read 16 bits from si made the debugger crash because it extended past the segment end. Now the debugger shows no actual values for that instruction, but should be fixed when it's about to execute and registers have differente values.


maximilien-noal (approved):

maximilien-noal:
@brunom I hit this crash once in a while but forgot about it, nice to see a fix for it! Thanks!

